PR #1, opened 2024-05-14
hello(world): initial action and CMakeLists

romintomasetti:
This PR introduces:
- a `devcontainer.json`
- an action that builds a ready-to-use image (that is pushed on the `GitHub` container registry) with
   * `Google Test` 1.14.0 compiled, installed
   * `Kokkos` 4.3.00 compiled, installed (with `OpenMP` backend)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,124 @@
+name: Build
+
+on:
+    push:
+        branches:
+            - main
+            - develop
+    pull_request:
+        branches:
+            - main
+            - develop
+
+env:
+    REGISTRY: ghcr.io
+
+jobs:
+
+    set-vars:
+        runs-on: [ubuntu-latest]
+        outputs:
+            CI_IMAGE : ${{ steps.common.outputs.CI_IMAGE }}
+        steps:
+            - name: Export common variables.
+              id  : common
+              run : |
+                  echo "CI_IMAGE=${{ env.REGISTRY }}/${{ github.repository }}/kokkos-utils:latest" >> $GITHUB_OUTPUT
+
+    build-image:
+        needs: [set-vars]
+        runs-on: [ubuntu-latest]
+        container:
+            image: docker:latest
+        permissions:
+            packages: write
+        steps:
+            - name: Checkout code.
+              uses: actions/checkout@v4
+
+            - name: Login to GitHub Container Registry.
+              uses: docker/login-action@v3
+              with:
+                  registry: ${{ env.REGISTRY }}
+                  username: ${{ github.actor }}
+                  password: ${{ secrets.GITHUB_TOKEN }}
+
+            - name: Build Docker image.
+              run : |
+                  apk add jq
+
+                  DOXYGEN_VERSION=$(jq .dependencies.doxygen.value version.json -j)
+                  GOOGLETEST_VERSION=$(jq .dependencies.googletest.value version.json -j)
+                  KOKKOS_VERSION=$(jq .dependencies.kokkos.value version.json -j)
+
+                  docker buildx build                                      \
+                      --pull                                               \
+                      --push                                               \
+                      --file dockerfile                                    \
+                      --platform linux/amd64                               \
+                      --tag ${{ needs.set-vars.outputs.CI_IMAGE }}         \
+                      --cache-from ${{ needs.set-vars.outputs.CI_IMAGE }}  \
+                      --build-arg BUILDKIT_INLINE_CACHE=1                  \
+                      --build-arg DOXYGEN_VERSION=${DOXYGEN_VERSION}       \
+                      --build-arg GOOGLETEST_VERSION=${GOOGLETEST_VERSION} \
+                      --build-arg KOKKOS_VERSION=${KOKKOS_VERSION}         \
+                      --build-arg KOKKOS_PRESET=OpenMP                     \
+                      --label "org.opencontainers.image.source=${{ github.repositoryUrl }}" \
+                      .
+
+    build-library:
+        needs: [set-vars, build-image]
+        runs-on: [ubuntu-latest]
+        container:
+            image: ${{ needs.set-vars.outputs.CI_IMAGE }}
+        steps:
+            - name: Checkout code.
+              uses: actions/checkout@v4
+
+            - name: Configure and build.
+              run : |
+                  cmake -S .    --preset=OpenMP
+                  cmake --build --preset=OpenMP
+
+    build-documentation:
+        needs: [set-vars, build-image]
+        runs-on: [ubuntu-latest]
+        container:
+            image: ${{ needs.set-vars.outputs.CI_IMAGE }}
+        steps:
+            - name: Checkout code.
+              uses: actions/checkout@v4
+
+            - name: Build Doxygen documentation.
+              run : |
+                  cmake -S . --preset=OpenMP
+                  cmake --build --preset=OpenMP --target=docs
+
+            - name: Upload Pages artifacts.
+              uses: actions/upload-pages-artifact@v3
+              with:
+                  path: build-with-OpenMP/docs/html
+
+    # The deploy job is heavily inspired from https://github.com/actions/deploy-pages.
+    deploy:
+        # The deployment only happens if the library can be built and the documentation generated.
+        needs: [build-library, build-documentation]
+        runs-on: [ubuntu-latest]
+
+        # The deployment only happens for 'main' or 'develop' branch.
+        if: ${{ contains(fromJSON('["refs/heads/main", "refs/heads/develop"]'), github.ref) }}
+
+        # Grant GITHUB_TOKEN the permissions required to make a Pages deployment
+        permissions:
+            pages: write      # to deploy to Pages
+            id-token: write   # to verify the deployment originates from an appropriate source
+
+        # Deploy to the github-pages environment
+        environment:
+            name: github-pages
+            url: ${{ steps.deployment.outputs.page_url }}
+
+        steps:
+            - name: Deploy to GitHub Pages.
+              id  : deployment
+              uses: actions/deploy-pages@v4
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,39 @@
+cmake_minimum_required(VERSION 3.12)
+
+#---- Read the version file.
+file(READ version.json VERSION_JSON)
+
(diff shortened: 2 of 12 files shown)


romintomasetti:
@maartenarnst The playground is nearly set !

.github/workflows/build.yml
```diff
@@ -0,0 +1,73 @@
+name: Build
+
+on:
```
maartenarnst, line 3: We could configure the repository to have "main"/"develop" branches, and require that PRs go into the develop branch. And later, we could also split the "build-image" and "build-library" into separate workflows, and then run "build-image" only when needed.

.github/workflows/build.yml
```diff
@@ -0,0 +1,73 @@
...
+        branches:
+            - main
+
+env:
+    REGISTRY: ghcr.io
+
+jobs:
+
+    set-env:
+        runs-on: [ubuntu-22.04]
```
maartenarnst: We could also do `ubuntu-latest`.

.github/workflows/build.yml
```diff
@@ -0,0 +1,73 @@
...
+    pull_request:
+        branches:
+            - main
+
+env:
+    REGISTRY: ghcr.io
+
+jobs:
+
+    set-env:
```
maartenarnst: A bit nicpicky :) But should this be named "set-vars" because we're setting an "output variable" in this case.

.github/workflows/build.yml
```diff
@@ -0,0 +1,73 @@
...
+            CI_IMAGE : ${{ steps.common.outputs.CI_IMAGE }}
+        steps:
+            - name: Export common variables.
+              id  : common
+              run : |
+                  echo "CI_IMAGE=${{ env.REGISTRY }}/${{ github.repository }}/kokkos-utils:latest" >> $GITHUB_OUTPUT
+
+    build-image:
+        needs: [set-env]
+        runs-on: [ubuntu-22.04]
```
maartenarnst: Could probably also be `ubuntu-latest`.

.github/workflows/build.yml
```diff
@@ -0,0 +1,73 @@
...
+            image: docker:latest
+        permissions:
+            packages: write
+        steps:
+            - name: Checkout code.
+              uses: actions/checkout@v4
+
+            - name: Log in to Docker registry.
+              run : |
+                  docker login -u ${{ github.actor}} -p ${{ secrets.GITHUB_TOKEN }} ${{ env.REGISTRY }}
```
maartenarnst: Reading 

- https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions

I saw that there are also "docker/login-action" and "docker/build-push-action" actions

- https://github.com/docker/login-action
- https://github.com/docker/build-push-action

It seems this could allow us to do these things within an `ubuntu-latest` image.

Something like:

```
name: Login to GitHub Container Registry
        uses: docker/login-action@v3
        with:
          registry: ${{ env.REGISTRY }}
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}
```

It seems the build-push-action works with buildx and all the parameters that we need.

.github/workflows/build.yml
```diff
@@ -0,0 +1,73 @@
...
+                      --cache-from ${{ needs.set-env.outputs.CI_IMAGE }}   \
+                      --build-arg BUILDKIT_INLINE_CACHE=1                  \
+                      --build-arg GOOGLETEST_VERSION=${GOOGLETEST_VERSION} \
+                      --build-arg KOKKOS_VERSION=${KOKKOS_VERSION}         \
+                      --build-arg KOKKOS_PRESET=OpenMP                     \
+                      .
+
+    build-library:
+        needs: [set-env, build-image]
+        runs-on: [ubuntu-22.04]
```
maartenarnst:
```suggestion
        runs-on: [ubuntu-latest]
```

CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+set(CMAKE_CXX_STANDARD_REQUIRED True CACHE BOOL "" FORCE)
+
+#---- Find Kokkos.
+find_package(
+    Kokkos
+    REQUIRED
+    ${KOKKOS_VERSION}
+)
+
+#---- Find google Test.
```
maartenarnst:
```suggestion
#---- Find Google Test.
```

dockerfile
```diff
@@ -0,0 +1,74 @@
+FROM ubuntu:24.04 as apt-requirements
+
+# Install APT requirements.
+RUN --mount=target=/requirements,type=bind,source=requirements <<EOF
```
maartenarnst, line 4: OK. That's nice!

dockerfile
```diff
@@ -0,0 +1,74 @@
...
+
+EOF
+
+# Compile Kokkos.
+FROM apt-requirements as compile-kokkos
+
+ARG KOKKOS_VERSION
+ARG KOKKOS_PRESET
+
+COPY kokkos.cmake.presets.json /kokkos.cmake.presets.json
```
maartenarnst: We could also create a `presets` directory, put the `kokkos.cmake.presets.json` there, and then use the `mount` like you did for google test.

dockerfile
```diff
@@ -0,0 +1,74 @@
...
+
+ARG GOOGLETEST_VERSION
+ARG KOKKOS_VERSION
+ARG KOKKOS_PRESET
+
+COPY --from=compile-google-test /opt/google-test-${GOOGLETEST_VERSION}         /opt/google-test-${GOOGLETEST_VERSION}
+COPY --from=compile-kokkos      /opt/kokkos-${KOKKOS_VERSION}/${KOKKOS_PRESET} /opt/kokkos-${KOKKOS_VERSION}/${KOKKOS_PRESET}
+
+ENV GTest_ROOT=/opt/google-test-${GOOGLETEST_VERSION}
+ENV Kokkos_ROOT=/opt/kokkos-${KOKKOS_VERSION}/${KOKKOS_PRESET}
```
maartenarnst, line 100: Reading

- https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry
- https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#labelling-container-images

we could also add a `org.opencontainers.image.source` label:

```
LABEL org.opencontainers.image.source=https://github.com/uliegecsm/kokkos-utils
```

(and perhaps other labels too)

maartenarnst: There is also a label `org.opencontainers.image.licenses`. So we should probably think about the license again :) First option is probably to use the same one as Kokkos?

maartenarnst (commented):
OK! Great! These are a few suggestions.

.github/workflows/build.yml
romintomasetti: Done :+1:

.github/workflows/build.yml
romintomasetti: Done :+1:

.github/workflows/build.yml
romintomasetti: Done :+1:

.github/workflows/build.yml
romintomasetti: Done for `login`.

.github/workflows/build.yml
romintomasetti: Done :+1:

CMakeLists.txt
romintomasetti: Done :+1:

dockerfile
romintomasetti: :+1:

dockerfile
romintomasetti: Well, in this case, the `COPY` does not hurt, because in the final image it will not be copied (only the `/opt/kokkos-<version>/<preset>` will be)...

.github/workflows/build.yml
romintomasetti: * I just created a `develop` branch, that is now the default branch. This PR now targets the `develop` branch.
* For the split of the CICD, I would do it in a followup :wink:

.github/workflows/build.yml
romintomasetti: I will not do it for the `build` part (yet).

dockerfile
romintomasetti: I've used the Apache v2 license, this is the closest I could find.

And done for adding the label to the image.

dockerfile
romintomasetti: I'll do as in https://github.com/serious-scaffold/ss-cpp/tree/master/cmake/presets. We'll store the presets that are not "the default one related to building a sane default version of our project" in `cmake/presets`.

romintomasetti:
@maartenarnst On my side, it's ready :rocket:

.github/workflows/build.yml
```diff
@@ -0,0 +1,124 @@
...
+                  cmake -S . --preset=OpenMP
+                  cmake --build --preset=OpenMP --target=docs
+
+            - name: Upload Pages artifacts.
+              uses: actions/upload-pages-artifact@v3
+              with:
+                  path: build-with-OpenMP/docs/html
+
+    # The deploy job is heavily inspired from https://github.com/actions/deploy-pages.
+    deploy:
```
maartenarnst, line 103: OK! That's nice!

dockerfile
```diff
@@ -0,0 +1,100 @@
...
+
+    apt update
+
+    apt --yes --no-install-recommends install $(grep -vE "^\s*#" /requirements/requirements.system.txt  | tr "\n" " ")
+
+    apt clean && rm -rf /var/lib/apt/lists/*
+
+EOF
+
+# Compile Doxygen.
```
maartenarnst, line 16: OK. Just wondering if we could rely on `apt` to install `Doxygen` for us? Because we're on `ubuntu:24.04`, the installed version would be `1.9.8`, which, would be good (?).

maartenarnst: Ah, OK, I see, in the version you would like `1.10.0`. But perhaps `1.9.8` would be fine?

romintomasetti: I guess 1.9.8 could be fine, but since we're going to use cutting edge C++ features, I prefer to stick with the most recent Doxygen release 😉

maartenarnst: OK to keep it this way!

maartenarnst (approved):
Thumps up for the first PR!

PR #2, opened 2024-05-15
Better doxygen + allow GPG sharing with host while inside `devcontainer`

romintomasetti:
## Summary

This PR:
1. Adds useful options to `Doxygen`.
2. Changes a bit the `devcontainer` logic to allow for local GPG keys to be shared within the container, thus allowing GPG signed commits.


maartenarnst (approved):
OK!

PR #4, opened 2024-05-29
add a single concept and a related test

romintomasetti:
## Summary

This PR starts to fill our `kokkos-utils` library. It adds one concept, and its test.

## Description

We should use this PR to discuss the structure of this repository, and how we will contribute to it in the near future.

### DIFF
--- a/include/concepts/View.hpp
+++ b/include/concepts/View.hpp
@@ -1,0 +1,15 @@
+#ifndef KOKKOS_UTILS_CONCEPTS_VIEW_HPP
+#define KOKKOS_UTILS_CONCEPTS_VIEW_HPP
+
+#include "Kokkos_View.hpp"
+
+namespace concepts
+{
+
+//! Concept to specify that a type is a @c Kokkos::View.
+template <typename T>
+concept View = Kokkos::is_view_v<T>;
+
+} // namespace concepts
+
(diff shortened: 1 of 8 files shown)


include/concepts/View.hpp
```diff
@@ -0,0 +1,15 @@
+#ifndef KOKKOS_UTILS_CONCEPTS_VIEW_HPP
```
maartenarnst, line 1: I wonder if we should repeat the name of the repository as a subdirectory of `include`. E.g., we'd have 

```
include/
    kokkos-utils/
        concepts/
            View.hpp
tests/
    Kokkos-utils/
        concepts/
            test_View.cpp
```

An advantage would be that if the include directory is included in another project, then, within that other project, one can write `#include "kokkos-utils/concepts/View.hpp"`. 

It would be like e.g. `googletest` that does `include/gtest/...` and we include like `#include "gtest/gtest.h"`.

In the `Kokkos` family, it seems `mdspan` does it this way, but the other subprojects like `kokkos-kernels`, `kokkos-comm`,... don't.

I think it would have an impact if we include `kokkos-utils` as a git submodule. But when we go through an install,  the install dirs can probably be easily reconfigured.

Another question is that if we start "small", we may not immediately need subfolders like `concepts`.

PR #7, opened 2024-06-20
atomics: insert op with atomic min

romintomasetti:
## Summary

This PR introduces the `atomics` namespace.

In this namespace, we now have `InsertOp.hpp`, a collection of helpers for inserting atomically.

In this `InsertOp.hpp`, we now have `InsertMin`, that can insert an element at a given index in a view, using `Kokkos::atomic_min`. It might be used with `Kokkos::UnorderedMap::insert` for instance.

## Description

Since the test `test_InsertOp.cpp` needs to allocate `Kokkos::View`s, we need to initialize `Kokkos`. Therefore, I introduced an object library with a `main.cpp` that initializes both `Google Test` and `Kokkos`.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -38,13 +38,21 @@ function(add_one_test)
             ${SOURCE_FILE}
     )
 
+    # Compile options.
+    target_compile_options(
+        ${EXECUTABLE_NAME}
+        PRIVATE
+            -Wall
+            -Wextra
+            -Werror
+    )
+
     # Link the executable to the library, GTest and Kokkos.
     target_link_libraries(
         ${EXECUTABLE_NAME}
         PRIVATE
             Kokkos::utils
-            GTest::gtest_main
-            Kokkos::kokkoscore
+            test_common
     )
 
     # Add the test to CTest.
@@ -54,6 +62,23 @@ function(add_one_test)
     )
 
 endfunction()
+
+# Common test library (object-like) which each test links to.
+add_library(test_common OBJECT)
+target_sources(
+    test_common
+    PRIVATE
+        main.cpp
+)
+target_link_libraries(
+    test_common
+    PUBLIC
+        GTest::gtest
+        Kokkos::kokkoscore
+)
+
+### TESTS : atomics ###
+add_subdirectory(atomics)
 
 ### TESTS : concepts ###
 add_subdirectory(concepts)
--- a/tests/atomics/test_InsertOp.cpp
+++ b/tests/atomics/test_InsertOp.cpp
@@ -1,0 +1,52 @@
+#include "gtest/gtest.h"
+
+#include "Kokkos_Core.hpp"
+
+#include "kokkos-utils/atomics/InsertOp.hpp"
+
+using execution_space = Kokkos::DefaultExecutionSpace;
+
+/**
+ * @file
+ *
+ * @addtogroup unittests
+ *
+ * **Atomic insertion**
+ *
+ * This group of tests check the behavior of our atomic insert operators.
+ */
+
+namespace Kokkos::utils::tests::atomics
+{
+
+//! @test Check that @ref Kokkos::utils::atomics::InsertMin works as expected.
+TEST(atomics, InsertMin)
+{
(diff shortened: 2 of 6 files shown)


tests/CMakeLists.txt
```diff
@@ -55,5 +54,22 @@ function(add_one_test)
 
 endfunction()
 
+# Common test library (object-like) to which each test links to.
```
maartenarnst:
```suggestion
# Common test library (object-like) which each test links to.
```

romintomasetti: :recycle:  :heavy_check_mark:

tests/atomics/test_InsertOp.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+    Kokkos::parallel_for(
+        "atomics::InsertMin",
+        Kokkos::RangePolicy<execution_space>(0, trials),
+        KOKKOS_LAMBDA(const typename execution_space::size_type trial)
+        {
+            insert_min.op(data, 0, trials - trial);
+        }
+    );
+}
```
maartenarnst:
```suggestion
TEST(atomics, InsertMin)
{
    using view_t = Kokkos::View<int*, execution_space>;

    const view_t data("data", 2);

    const utils::atomics::InsertMin insert_min {};

    Kokkos::parallel_for(
        "atomics::InsertMin",
        Kokkos::RangePolicy<execution_space>(0, 1),
        KOKKOS_LAMBDA(const typename execution_space::size_type /* */)
        {
            insert_min.op(data, 0, -3);
            insert_min.op(data, 1,  2);
        }
    );
    Kokkos::fence("atomics_InsertMin_Test: fence after parallel_for operation");    

    const auto data_h = Kokkos::create_mirror_view_and_copy(Kokkos::DefaultHostExecutionSpace{}, data);

    ASSERT_EQ(data_h(0), -3);
    ASSERT_EQ(data_h(1),  0);    
}
```
(to test the outcome of the operation)

romintomasetti: OK for yu ?

tests/atomics/test_InsertOp.cpp
```diff
@@ -0,0 +1,52 @@
...
+//! @test Check that @ref Kokkos::utils::atomics::InsertMin works as expected.
+TEST(atomics, InsertMin)
+{
+    using view_t = Kokkos::View<int*, execution_space>;
+
+    constexpr size_t trials = 150;
+
+    const execution_space space {};
+
+    view_t data(Kokkos::view_alloc(space, "data"), 2);
```
maartenarnst: This view can probably be `const`.

tests/atomics/test_InsertOp.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+    const utils::atomics::InsertMin insert_min {};
+
+    Kokkos::parallel_for(
+        "atomics::InsertMin",
+        Kokkos::RangePolicy<execution_space>(space, 0, trials),
+        KOKKOS_LAMBDA(const typename execution_space::size_type trial)
+        {
+            insert_min.op(data, 0, trials - trial - 2);
```
maartenarnst, line 40: To avoid having to think about mixing signed and unsigned types, we could also declare `trial` as an `int` and do `trial - 2`. but not sure if there's an issue here.

tests/atomics/test_InsertOp.cpp
romintomasetti: Indeed, and done.

tests/atomics/test_InsertOp.cpp
romintomasetti: It seems that the compiler is fine, even with `-Wall -Wextra -Werror`.

But I've changed to `int` still.

maartenarnst: OK!

tests/atomics/test_InsertOp.cpp
maartenarnst: 🟢 !

maartenarnst (approved):
OK! Nice addition 👍

PR #8, opened 2024-06-20
cicd: add schedule to cope for updated base image

romintomasetti:
## Summary

This PR introduces a scheduled action.

I followed best practice and set the *cron* start at end of night, and not at the start of an hour to avoid high load times of GitHub runners.

This schedule will basically do not much if our base images from `Docker` hub haven't changed.


maartenarnst (approved):
OK! Let's repeat it every Wednesday at 5h42!

PR #11, opened 2024-06-20
cicd(backend): add `Cuda`

romintomasetti:
## Summary

This PR adds support for `Cuda`.

- [x] See how we can run the tests. If it's too hard to get a runner with a `Cuda` GPU, let's just compile the code.

### DIFF
--- a/.devcontainer/dockerfile
+++ b/.devcontainer/dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/uliegecsm/kokkos-utils/kokkos-utils:latest
+FROM ghcr.io/uliegecsm/kokkos-utils/kokkos-utils:gcc-OpenMP
 
 # Install gpg and gnupg2 allows the container to use GPG keys to sign commits.
 # See also:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
             - name: Export common variables.
               id  : common
               run : |
-                  echo "CI_IMAGE=${{ env.REGISTRY }}/${{ github.repository }}/kokkos-utils:latest" >> $GITHUB_OUTPUT
+                  echo "CI_IMAGE=${{ env.REGISTRY }}/${{ github.repository }}/kokkos-utils" >> $GITHUB_OUTPUT
 
     build-image:
         needs: [set-vars]
@@ -36,6 +36,19 @@ jobs:
             image: docker:latest
         permissions:
             packages: write
+        strategy:
+            fail-fast: false
+            matrix:
+                include:
+                    - backend        : Cuda
+                      compiler_family: nvidia
+                      base_image     : nvidia/cuda:12.3.2-devel-ubuntu22.04
+                      platforms      : linux/amd64
+                      kokkos_arch    : VOLTA70
+                    - backend        : OpenMP
+                      compiler_family: gcc
+                      base_image     : ubuntu:24.04
+                      platforms      : linux/amd64,linux/arm64
         steps:
             - name: Checkout code.
               uses: actions/checkout@v4
@@ -44,9 +57,10 @@ jobs:
               run: |
                   apk add jq
 
-                  echo "DOXYGEN_VERSION=$(jq .dependencies.doxygen.value version.json -j)" >> $GITHUB_ENV
+                  echo "CMAKE_VERSION=$(     jq .dependencies.cmake.value      version.json -j)" >> $GITHUB_ENV
+                  echo "DOXYGEN_VERSION=$(   jq .dependencies.doxygen.value    version.json -j)" >> $GITHUB_ENV
                   echo "GOOGLETEST_VERSION=$(jq .dependencies.googletest.value version.json -j)" >> $GITHUB_ENV
-                  echo "KOKKOS_VERSION=$(jq .dependencies.kokkos.value version.json -j)" >> $GITHUB_ENV
+                  echo "KOKKOS_VERSION=$(    jq .dependencies.kokkos.value     version.json -j)" >> $GITHUB_ENV
 
             - name: Set up QEMU.
               uses: docker/setup-qemu-action@v3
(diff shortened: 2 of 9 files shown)


docker/dockerfile
```diff
@@ -7,6 +9,14 @@ RUN --mount=target=/requirements,type=bind,source=requirements <<EOF
 
     apt update
 
+    # Be sure to install a recent CMake, see also https://apt.kitware.com/.
```
romintomasetti: Use something like
```docker
If not $(apt --yes --no-install-recommends satisfy "cmake (>=3.23)" then
   Add kitware
   Apt install cmake
Endif
```

romintomasetti: Or simply get the binaries at https://cmake.org/download/, it's gonna be easier.

.github/workflows/build.yml
```diff
@@ -65,49 +76,59 @@ jobs:
               uses: docker/build-push-action@v5
               with:
                   context: .
-                  platforms: linux/amd64,linux/arm64
-                  push: ${{ github.ref == 'refs/heads/develop' }}
+                  platforms: ${{ matrix.platforms }}
+                  push: ${{ github.ref == 'refs/heads/develop' || true }}
```
romintomasetti: To be reverted ?

docker/dockerfile
romintomasetti: OK, done.

romintomasetti:
@maartenarnst Ready for review!

.github/workflows/build.yml
maartenarnst: Yes, best to be reverted, because it I think it won't work when someone does a PR from a fork.

.github/workflows/build.yml
```diff
@@ -36,6 +36,17 @@ jobs:
             image: docker:latest
         permissions:
             packages: write
+        strategy:
+            fail-fast: false
+            matrix:
+                include:
+                    - backend    : Cuda
```
maartenarnst: Let's add the compiler family, currently `gcc` for the `OpenMP` backend and `nvidia` for the `Cuda` backend. The tags for the images could then become `gcc-OpenMP` and `nvidia-Cuda`.

maartenarnst: I'd modify the preset names like this too, i.e. `gcc-OpenMP` and `nvidia-Cuda`.

.github/workflows/build.yml
```diff
@@ -65,49 +77,62 @@ jobs:
...
-    build-library:
+    build-library-and-test:
         needs: [set-vars, build-image]
         runs-on: [ubuntu-latest]
+        strategy:
+            fail-fast: false
+            matrix:
+                include:
+                    - backend: Cuda
+                      tests  : false
```
maartenarnst: The name for the flag could also be `run_tests` to emphasise that we compile the tests, but it's controlling whether we run the tests.

.github/workflows/build.yml
romintomasetti: OK, let's go! :car:

.github/workflows/build.yml
romintomasetti: OK, let's go !

.devcontainer/dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/uliegecsm/kokkos-utils/kokkos-utils:latest
+FROM ghcr.io/uliegecsm/kokkos-utils/kokkos-utils:OpenMP
```
maartenarnst:
```suggestion
FROM ghcr.io/uliegecsm/kokkos-utils/kokkos-utils:gcc-OpenMP
```

.github/workflows/build.yml
maartenarnst: Still a todo :)

.github/workflows/build.yml
```diff
@@ -65,49 +79,64 @@ jobs:
...
               run : |
-                  ctest --preset=OpenMP
+                  ctest --preset=${{ matrix.compiler_family }}-${{ matrix.backend }}
 
     build-documentation:
         needs: [set-vars, build-image]
         runs-on: [ubuntu-latest]
         container:
-            image: ${{ needs.set-vars.outputs.CI_IMAGE }}
+            image: ${{ needs.set-vars.outputs.CI_IMAGE }}:OpenMP
```
maartenarnst:
```suggestion
            image: ${{ needs.set-vars.outputs.CI_IMAGE }}:gcc-OpenMP
```

.github/workflows/build.yml
```diff
@@ -65,49 +79,64 @@ jobs:
...
         steps:
             - name: Checkout code.
               uses: actions/checkout@v4
 
             - name: Build Doxygen documentation.
               run : |
-                  cmake -S . --preset=OpenMP
+                  cmake -S . --preset=OpenMP --warn-uninitialized -Werror=dev
```
maartenarnst:
```suggestion
                  cmake -S . --preset=gcc-OpenMP --warn-uninitialized -Werror=dev
```

.devcontainer/dockerfile
romintomasetti: :heavy_check_mark:

.github/workflows/build.yml
romintomasetti: :heavy_check_mark:

.github/workflows/build.yml
romintomasetti: :heavy_check_mark:

.github/workflows/build.yml
romintomasetti: Yes, but if I revert now, I won't have the correct image when pulling at compile stage...

.github/workflows/build.yml
romintomasetti: Done.

.github/workflows/build.yml
romintomasetti: Done.

maartenarnst (dismissed):
OK!

.github/workflows/build.yml
romintomasetti: Reverted.

maartenarnst (approved):
OK!

PR #16, opened 2024-09-02
feat: add extensions to `Kokkos::Impl::type_list`

maartenarnst:
## Summary

This PR adds extensions to `Kokkos::Impl::type_list`.

### DIFF
--- a/include/kokkos-utils/impl/type_list.hpp
+++ b/include/kokkos-utils/impl/type_list.hpp
@@ -1,0 +1,78 @@
+#ifndef KOKKOS_UTILS_IMPL_TYPE_LIST_HPP
+#define KOKKOS_UTILS_IMPL_TYPE_LIST_HPP
+
+#include "impl/Kokkos_Utilities.hpp"
+
+/**
+ * @file
+ *
+ * This file provides extensions to the @c Kokkos::Impl::type_list struct in @c Kokkos_Utilities.hpp.
+ */
+
+namespace Kokkos::utils::impl
+{
+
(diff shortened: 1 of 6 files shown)


include/kokkos-utils/impl/type_list.hpp
```diff
@@ -0,0 +1,78 @@
...
+//! @name Get the number of types in a @c Kokkos::Impl::type_list.
+///@{
+template <typename>
+struct TypeListSize;
+
+template <typename... Ts>
+struct TypeListSize<Kokkos::Impl::type_list<Ts...>> : std::integral_constant<size_t, sizeof...(Ts)> {};
+
+template <typename T>
+inline constexpr size_t type_list_size_v = TypeListSize<T>::value;
```
romintomasetti, line 24: Just wondering if constexpr implies inline already.

romintomasetti (changes requested):

include/kokkos-utils/impl/type_list.hpp
maartenarnst: It's different for functions and variables. It's only for static variables that `constexpr` implies `inline`. It's summarised e.g. in the box at the top of this link:
- https://en.cppreference.com/w/cpp/language/inline

Btw, the `type_list_size_v` and `type_list_at_t` are directly taken from example code for `std::tuple`. I'll add that to the description of the PR.

romintomasetti: I guess it's fine like that.

romintomasetti (approved):

PR #20, opened 2024-09-10
view(slice): take a subview without specifying all trailing Kokkos::ALL

romintomasetti:
## Summary

This PR brings the `slice` utility.

Drive-by changes:
- `rank()` as suggested in `Kokkos` documentation.
- Titles of Doxygen sections.

### DIFF
--- a/include/kokkos-utils/view/slice.hpp
+++ b/include/kokkos-utils/view/slice.hpp
@@ -1,0 +1,78 @@
+#ifndef KOKKOS_UTILS_VIEW_SLICE_HPP
+#define KOKKOS_UTILS_VIEW_SLICE_HPP
+
+#include "kokkos-utils/concepts/View.hpp"
+
+/**
+ * @file
+ *
+ * This file is a collection of helper functions similar to what can be found in
+ * the standard @c mdspan header, related to retrieving slices of a @c Kokkos::View.
+ *
+ * References:
+ *  - https://en.cppreference.com/w/cpp/container/mdspan
+ *  - https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2630r4.html
+ */
+
+namespace Kokkos::utils::view
+{
+
+namespace impl
+{
+
+template <typename ViewType, typename... Indices, size_t... AllsIndices>
+KOKKOS_FUNCTION
+constexpr auto slice(ViewType&& view, Indices&&... indices, std::index_sequence<AllsIndices...>)
+{
+    return Kokkos::subview(
+        std::forward<ViewType>(view),
+        std::forward<Indices>(indices)...,
+        std::get<AllsIndices>(std::array<Kokkos::ALL_t, sizeof...(AllsIndices)>{})...
+    );
+}
+
+} // namespace impl
+
+/**
+ * @brief Get a subview, given the first @p indices. The rest is filled with @c Kokkos::ALL.
+ *
+ * @note Though similar to @c std::submdspan, we think it it OK not to provide the remaining
+ *       slice specifiers if they are all @c Kokkos::ALL iif the user specifies what is the expected rank
+ *       of the input view. This ensures that the user will not face unexpected behaviors.
+ *       Therefore, we allow the following:
+ *       @code
+ *       auto sliced = slice<4>(view_of_rank_4, 1, 2);
+ *       @endcode
+ *       which is strictly equivalent to:
+ *       @code
+ *       auto sliced = subview(view_of_rank_4, 1, 2, Kokkos::ALL, Kokkos::ALL);
+ *       @endcode
+ */
+template <size_t Rank, typename ViewType, typename... Indices>
+requires concepts::ViewOfRank<std::remove_reference_t<ViewType>, Rank>
+KOKKOS_FUNCTION
+constexpr auto slice(ViewType&& view, Indices&&... indices)
+{
+    constexpr auto size = Rank - sizeof...(Indices);
+    return impl::slice<ViewType, Indices...>(
+        std::forward<ViewType>(view),
+        std::forward<Indices>(indices)...,
+        std::make_index_sequence<size>{}
+    );
+}
+
+//! @overload
+template <typename ViewType, typename... Indices>
+requires concepts::ViewOfRank<std::remove_reference_t<ViewType>, sizeof...(Indices)>
+KOKKOS_FUNCTION
+constexpr auto slice(ViewType&& view, Indices&&... indices)
+{
+    return slice<std::remove_reference_t<ViewType>::rank()>(
+        std::forward<ViewType>(view),
+        std::forward<Indices>(indices)...
+    );
+}
+
+} // namespace Kokkos::utils::view
+
+#endif // KOKKOS_UTILS_VIEW_SLICE_HPP
--- a/tests/view/test_extents.cpp
+++ b/tests/view/test_extents.cpp
@@ -12,7 +12,8 @@ using execution_space = Kokkos::DefaultExecutionSpace;
  *
  * @addtogroup unittests
  *
- * **Get extents of a @c Kokkos::View**
+ * Get extents of a @c Kokkos::View
+ * --------------------------------
  *
  * This group of tests check the behavior of our helpers related to @c Kokkos::View that can be
  * found in @ref extents.hpp.
@@ -22,6 +23,14 @@ namespace Kokkos::utils::tests::view
 {
 
 constexpr size_t dim_1 = 5, dim_2 = 3, dim_3 = 6, dim_4 = 9;
+
+//! @test Check that @ref view::extents works as expected for a rank-0 @c Kokkos::View.
+TEST(view, extents_rank_0)
+{
+    const Kokkos::View<double, execution_space> view_0("rank-0 view");
+
+    static_assert(utils::view::extents(view_0) == Kokkos::Array<size_t, 0>{});
+}
 
 //! @test Check that @ref view::extents works as expected for a rank-1 @c Kokkos::View.
 TEST(view, extents_rank_1)
--- a/tests/view/test_slice.cpp
+++ b/tests/view/test_slice.cpp
@@ -1,0 +1,119 @@
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+#include "Kokkos_Core.hpp"
+
+#include "kokkos-utils/view/extents.hpp"
+#include "kokkos-utils/view/slice.hpp"
+
+using execution_space = Kokkos::DefaultExecutionSpace;
+
+/**
+ * @file
+ *
+ * @addtogroup unittests
+ *
+ * Get slices of a @c Kokkos::View
+ * -------------------------------
+ *
+ * This group of tests check the behavior of our helpers related to @c Kokkos::View that can be
+ * found in @ref slice.hpp.
+ */
+
+namespace Kokkos::utils::tests::view
+{
+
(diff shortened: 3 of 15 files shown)


include/kokkos-utils/view/slice.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Kokkos::utils::view
+{
+
+namespace impl
+{
+
+template <typename ViewType, typename TupleA, size_t... A, typename... Indices>
+KOKKOS_FUNCTION
+constexpr auto slice(ViewType&& view, TupleA&& alls, std::index_sequence<A...>, Indices&&... indices) {
```
maartenarnst: The `typename ViewType` can probably use the concept. We'll also have to think about whether we can find better names for `TupleA` and especially `A`.

maartenarnst: Perhaps something like an abbreviated `Inds` with a good suffix for each could be used for `A` and `Indices`?

maartenarnst: I'm also wondering about the `constexpr`. It looks like in `Kokkos`, the `subview` function is not marked `constexpr`:

- https://github.com/kokkos/kokkos/blob/53ceb4c2d2095fa8173f95652f242f82ca982f12/core/src/View/Kokkos_ViewLegacy.hpp#L1408-L1418

maartenarnst: Another thought may be whether we should pass the indices by universal reference. Its's just integers, so they may not support move operators. Perhaps just passing by value would be fine for those. May be worth taking a look in `mdspan` to see how it was passed there.

maartenarnst: If I may one more 😄. The `TupleA` could also become `ArrayA` because we're passing an `std::array`.

include/kokkos-utils/view/slice.hpp
```diff
@@ -0,0 +1,78 @@
...
+    constexpr auto size = Rank - sizeof...(Indices);
+    return impl::slice(
+        std::forward<ViewType>(view),
+        std::array<Kokkos::ALL_t, size>{},
+        std::make_index_sequence<size>{},
+        std::forward<Indices>(indices)...
+    );
+}
+
+//! @overload
```
maartenarnst, line 64: OK. That's nice how the two functions work together.

tests/view/test_extents.cpp
```diff
@@ -23,6 +24,14 @@ namespace Kokkos::utils::tests::view
 
 constexpr size_t dim_1 = 5, dim_2 = 3, dim_3 = 6, dim_4 = 9;
 
+//! @test Check that @ref view::extents works as expected for a rank-0 @c Kokkos::View.
+TEST(view, extents_rank_0)
+{
+    const Kokkos::View<double, execution_space> view_0("rank-0 view");
+
+    ASSERT_EQ(utils::view::extents(view_0), (Kokkos::Array<size_t, 0>{}));
```
maartenarnst: OK. May be worth a try to make it a `static_assert`?

tests/view/test_slice.cpp
```diff
@@ -0,0 +1,119 @@
...
+ */
+
+namespace Kokkos::utils::tests::view
+{
+
+#define CHECK_THAT(__slice__, __rank__, __extents__)       \
+    {                                                      \
+        const auto tmp = __slice__;                        \
+        ASSERT_EQ(tmp.rank(), __rank__);                   \
+        ASSERT_EQ(utils::view::extents(tmp), __extents__); \
```
maartenarnst, line 30: Same for those :) May be worth trying with `static_assert`.

maartenarnst: I'm also wondering how it works with the dynamic extents. May be interesting to see whether the `static_assert` works with dynamic extents in particular.

include/kokkos-utils/view/slice.hpp
romintomasetti: :+1:

tests/view/test_slice.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    Kokkos::parallel_reduce(
+        Kokkos::RangePolicy<execution_space>(0, 1),
+        KOKKOS_LAMBDA(const int, bool& result)
+        {
+            const auto subview = Kokkos::subview      (view, 0, Kokkos::ALL, Kokkos::ALL, Kokkos::ALL);
+            const auto slice   = utils::view::slice<4>(view, 0);
+            result = (
+                subview.size() == expected_size && subview.rank == 3 &&
+                slice  .size() == expected_size && slice  .rank == 3 &&
```
maartenarnst: Should it be `rank()`?

tests/view/test_extents.cpp
romintomasetti: OK, done.

tests/view/test_slice.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    CHECK_THAT(utils::view::slice<4>(view_4_static, 0, Kokkos::ALL), 3, (Kokkos::Array{dim_2, dim_3, dim_4}));
+
+    CHECK_THAT(utils::view::slice(view_4_static, 0, 0, 0, 0), 0, (Kokkos::Array<size_t, 0>{}));
+
+    CHECK_THAT(utils::view::slice<4>(view_4_static, Kokkos::make_pair(0, 2)), 4, (Kokkos::Array<size_t, 4>{2, dim_2, dim_3, dim_4}));
+}
+
+template <concepts::ViewOfRank<4> ViewType>
+bool slice_on_device(const ViewType& view, const size_t expected_size)
```
maartenarnst: Could also become `test_slice_on_device` with a convention to start such functions with `test_`? May be worth checking in `Kokkos` how such functions are named.

tests/view/test_slice.cpp
romintomasetti: I won't work, not doing what you suggest.

tests/view/test_slice.cpp
romintomasetti: Indeed, done.

tests/view/test_slice.cpp
romintomasetti: Suggestion applied, haven't checked on the `Kokkos` side though.

include/kokkos-utils/view/slice.hpp
romintomasetti: 1. `ViewType` could be constrained with a concept, but considering it is in `impl`, I will not do it. It should reduce compiler pressure probably.
2. `constexpr` is there for the future. When `View` will be a wrapper around `std::mdspan` I expect this to also be used in compile-time expressions.
3. Indices are integers, tuple-like stuff or `Kokkos::ALL`. So I'll keep the forwarding.
4. I'll keep `AllsValues` and `AllsIndices`.

Please have a look at the new version.

include/kokkos-utils/view/slice.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Kokkos::utils::view
+{
+
+namespace impl
+{
+
+template <typename ViewType, typename AllsValues, size_t... AllsIndices, typename... Indices>
+KOKKOS_FUNCTION
+constexpr auto slice(ViewType&& view, AllsValues&& alls, std::index_sequence<AllsIndices...>, Indices&&... indices) {
```
maartenarnst: Just for the naming :) We could also rename `Indices&&... indices` as `SliceArgs&&... args` to reuse the names from the submdspan paper. It would also help see that it's not necessarily integers, but it can also be pairs, alls, ...

maartenarnst: After playing a bit with the code, I also have two variants to propose. The first variant would be to create the `alls` on the fly as needed, so it would avoid passing it as a parameter. We would have:

```
namespace impl {

template <typename ViewType, typename... Indices, size_t... AllsIndices>
KOKKOS_FUNCTION
constexpr auto slice(ViewType&& view, Indices&&... indices, std::index_sequence<AllsIndices...>) {    
    return Kokkos::subview(
        std::forward<ViewType>(view),
        std::forward<Indices>(indices)...,
        std::get<AllsIndices>(std::array<Kokkos::ALL_t, sizeof...(AllsIndices)>{})...
    );
}

} // namespace impl

template <size_t Rank, typename ViewType, typename... Indices>
requires concepts::ViewOfRank<std::remove_reference_t<ViewType>, Rank>
KOKKOS_FUNCTION
constexpr auto slice(ViewType&& view, Indices&&... indices)
{
    constexpr auto size = Rank - sizeof...(Indices);
    return impl::slice<ViewType, Indices...>(
        std::forward<ViewType>(view),
        std::forward<Indices>(indices)...,
        std::make_index_sequence<size>{}
    );
}
```

The second variant would replace the helper function in the `impl` namespace with a lambda. So the advantage would be that the code would be shorter. We'd have:

```
template <size_t Rank, typename ViewType, typename... Indices>
requires concepts::ViewOfRank<std::remove_reference_t<ViewType>, Rank>
KOKKOS_FUNCTION
constexpr auto slice(ViewType&& view, Indices&&... indices)
{
    constexpr auto size = Rank - sizeof...(Indices);
    return [view = std::forward<ViewType>(view), ... indices = std::forward<Indices>(indices)] 
        <size_t... AllsIndices>(std::index_sequence<AllsIndices...>) {
            return Kokkos::subview(
                std::forward<ViewType>(view),
                std::forward<Inds>(indices)...,
                std::get<AllsIndices>(std::array<Kokkos::ALL_t, size>{})...
            );
        }(std::make_index_sequence<size>{});
}
```

Up to you. We can keep the current version, or one of these variants.

include/kokkos-utils/view/slice.hpp
maartenarnst: OK. Just for the use of the `constexpr` for the future, I'm not sure what would be best. Another option would be declare the function without `constexpr` here, and then have a note/fixme in the docstring to say that we expect it to become `constexpr` at some point. An advantage would be that we can keep better track of what's `constexpr` and what's not.

maartenarnst:
OK. I've gone over the PR once more :) Another detail may be that we may want to have some policy for increasing the version, e.g. increase the minor version by one when we add a new feature?

tests/view/test_slice.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+#include "kokkos-utils/view/extents.hpp"
+#include "kokkos-utils/view/slice.hpp"
+
+using execution_space = Kokkos::DefaultExecutionSpace;
+
+/**
+ * @file
+ *
+ * @addtogroup unittests
```
maartenarnst, line 14: We noticed that it may not necessarily work to set both `@file` and `@addtogroup`. Let's remove `@file` here and to a search replace in the repository to solve similar cases?

include/kokkos-utils/view/slice.hpp
romintomasetti: I'll keep the `constexpr`.

include/kokkos-utils/view/slice.hpp
romintomasetti: The second one is too hard to read.

romintomasetti: Done.

tests/view/test_slice.cpp
romintomasetti: I will not do this here :wink:

maartenarnst (approved):
OK!

PR #21, opened 2024-09-18
cmake(bug): kokkos-utils target must depend on kokkos core target

romintomasetti:
## Summary

The `KokkosUtils` target *must* depend on `Kokkos::kokkoscore`. Otherwise, the `CMake` dependency tree is not correct.


maartenarnst (approved):
OK!

PR #31, opened 2025-03-18
cmake: check documentation list sizes

romintomasetti:
## Summary

Check how many files we have in the documentation.


maartenarnst (approved):
Perfect!

PR #32, opened 2025-03-19
cmake: install

romintomasetti:
## Summary

I need to be able to fetch this repository and install it. I don't want to integrate it as a submodule.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -131,6 +131,16 @@ jobs:
               run : |
                   ctest --preset=${{ matrix.compiler_family }}-${{ matrix.backend }}
 
+            - name: Install.
+              run : |
+                  bash tests/install.sh ${{ matrix.compiler_family }}-${{ matrix.backend }}
+
+            - name: Find package against install.
+              run : |
+                  test -d include/
+                  rm -rf include/
+                  bash tests/find_package.sh
+
     build-documentation:
         needs: [set-vars, build-image]
         runs-on: [ubuntu-latest]
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,57 +49,62 @@ endif()
 kokkos_check(DEVICES "${EXPECTED_Kokkos_DEVICES}")
 
 #---- Build the Kokkos::utils library.
-add_library(KokkosUtils INTERFACE)
-add_library(Kokkos::utils ALIAS KokkosUtils)
+add_library(kokkosutils INTERFACE)
+add_library(Kokkos::utils ALIAS kokkosutils)
 
 target_sources(
-    KokkosUtils
+    kokkosutils
     INTERFACE
-        include/kokkos-utils/atomics/InsertOp.hpp
+        FILE_SET headers
+        TYPE HEADERS
+        BASE_DIRS include/kokkos-utils
+        FILES
+            include/kokkos-utils/atomics/InsertOp.hpp
 
-        include/kokkos-utils/callbacks/EventInProfileSectionRegexMatcher.hpp
-        include/kokkos-utils/callbacks/EventRegexMatcher.hpp
-        include/kokkos-utils/callbacks/Events.hpp
-        include/kokkos-utils/callbacks/Helpers.hpp
-        include/kokkos-utils/callbacks/Listener.hpp
-        include/kokkos-utils/callbacks/Manager.hpp
-        include/kokkos-utils/callbacks/Matcher.hpp
-        include/kokkos-utils/callbacks/RecorderListener.hpp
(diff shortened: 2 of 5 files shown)


tests/install.sh
```diff
@@ -0,0 +1,15 @@
+set -ex
+
+CMAKE_PRESET=$1
+CMAKE_INSTALL_PREFIX=$2
+
+echo "> Running install test for CMake preset ${CMAKE_PRESET}. Installation is expected to happen in ${CMAKE_INSTALL_PREFIX}."
+
+cmake --build --preset=${CMAKE_PRESET} -j8 --target=install
```
maartenarnst: I wonder if we could do here 

```
cmake --install build-with-${CMAKE_PRESET} -j8 --prefix ${CMAKE_INSTALL_PREFIX}
```

to avoid that we have to add `-DCMAKE_INSTALL_PREFIX=/opt/test-folder-for-install` in the configure and build step in the GitHub workflow.

.github/workflows/build.yml
```diff
@@ -123,14 +123,24 @@ jobs:
 
             - name: Configure and build.
               run : |
-                  cmake -S .    --preset=${{ matrix.compiler_family }}-${{ matrix.backend }} --warn-uninitialized -Werror=dev
+                  cmake -S .    --preset=${{ matrix.compiler_family }}-${{ matrix.backend }} --warn-uninitialized -Werror=dev -DCMAKE_INSTALL_PREFIX=/opt/test-folder-for-install
                   cmake --build --preset=${{ matrix.compiler_family }}-${{ matrix.backend }}
 
+            - name: Install.
+              run : |
+                  bash tests/install.sh ${{ matrix.compiler_family }}-${{ matrix.backend }} /opt/test-folder-for-install
```
maartenarnst: Let's move it to the step after `Test` to make it "grouped" with the last step?

maartenarnst: The two could also become a single step.

docs/CMakeLists.txt
```diff
@@ -31,7 +31,9 @@ set(DOXYGEN_GENERATE_TAGFILE "${CMAKE_CURRENT_BINARY_DIR}/html/${CMAKE_PROJECT_N
 set(DOXYGEN_TAGFILES "${KOKKOS_TAG_FILE}")
 
 #---- Extract files for documentation.
-get_target_property(KokkosUtils_TARGET_FILES_FOR_DOC KokkosUtils INTERFACE_SOURCES)
+get_target_property(KokkosUtils_TARGET_FILES_FOR_DOC kokkosutils INTERFACE_HEADER_SETS)
+get_target_property(KokkosUtils_TARGET_FILES_FOR_DOC kokkosutils HEADER_SET_${KokkosUtils_TARGET_FILES_FOR_DOC})
```
maartenarnst, line 37: Should we use two different variables names?

.github/workflows/build.yml
```diff
@@ -123,14 +123,24 @@ jobs:
...
+
             - name: Test.
               if  : ${{ matrix.run_tests }}
               run : |
                   ctest --preset=${{ matrix.compiler_family }}-${{ matrix.backend }}
 
+            - name: Find package against install.
+              run : |
+                  test -d include/
+                  rm -rf include/
```
maartenarnst, line 141: Still needed?

CMakeLists.txt
```diff
@@ -48,59 +48,68 @@ endif()
...
 target_link_libraries(
-    KokkosUtils
+    kokkosutils
     INTERFACE
         Kokkos::kokkoscore
 )
 
+#---- Install.
+install(TARGETS kokkosutils EXPORT KokkosUtilsConfig FILE_SET headers DESTINATION include/kokkos-utils)
+install(EXPORT KokkosUtilsConfig DESTINATION lib/cmake/KokkosUtils NAMESPACE Kokkos::)
```
maartenarnst, line 120: Did you mean to write `NAMESPACE Kokkos::`?

tests/install.sh
romintomasetti: :heavy_check_mark:

.github/workflows/build.yml
romintomasetti: Kept it 2 steps.

docs/CMakeLists.txt
romintomasetti: Not needed, I did it on purpose.

.github/workflows/build.yml
romintomasetti: It's a safety measure to ensure that only the installed files exist.

CMakeLists.txt
romintomasetti: Yes.

maartenarnst (approved):
OK! 🟢

PR #34, opened 2025-03-21
docker: add clang image

romintomasetti:
## Summary

This PR adds a `clang`-based image.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,7 +41,7 @@ jobs:
             matrix:
                 include:
                     - backend        : Cuda
-                      compiler_family: nvidia
+                      compiler_family: gcc
                       base_image     : nvidia/cuda:12.8.1-devel-ubuntu24.04
                       platforms      : linux/amd64
                       kokkos_arch    : VOLTA70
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -80,7 +83,7 @@ jobs:
               with:
                   context: .
                   platforms: ${{ matrix.platforms }}
-                  push: ${{ github.ref == 'refs/heads/develop' }}
+                  push: true #${{ github.ref == 'refs/heads/develop' }}
```
romintomasetti: Revert this before merging.

docker/dockerfile
```diff
@@ -2,15 +2,41 @@ ARG BASE_IMAGE
...
     set -ex
 
-    apt update
+    python3 -m pip install git+https://github.com/uliegecsm/system-helpers
 
-    apt --yes --no-install-recommends install $(grep -vE "^\s*#" /requirements/requirements.system.txt  | tr "\n" " ")
+    if [ "${COMPILER_FAMILY}" = "gcc" ];then
+        COMPILER_FAMILY_REQUIREMENTS=g++
+    elif [ "${COMPILER_FAMILY}" = "clang" ];then
+        COMPILER_FAMILY_REQUIREMENTS="clang clang-tidy libomp-dev"
```
maartenarnst, line 31: Just wondering if we need `libstdc++-dev` for the `gcc` standard library in this case?

romintomasetti: Not needed in this case, we use defaulted versions of everything.

.github/workflows/build.yml
romintomasetti: :heavy_check_mark:

maartenarnst (approved):
OK! Great!

PR #39, opened 2025-04-13
feat(`cicd`): add `rocm` image and build workflows

maartenarnst:
## Summary

This PR adds `rocm` to the `build-image` and `build-library-and-test` workflows.

As a drive-by change, the `Kokkos` version is updated to 4.6.00.


.github/workflows/build.yml
```diff
@@ -139,6 +147,13 @@ jobs:
...
                   test -d include/
                   rm -rf include/
+
+                  if [ ${{ matrix.compiler_family }} = "clang" ];then
+                    export CXX=clang++
+                  elif [ ${{ matrix.compiler_family }} = "rocm" ];then
+                    export CXX=hipcc
+                  fi
+
                   bash tests/find_package.sh
```
romintomasetti: I would rather have it as an argument of the `find_package.sh` script.

maartenarnst: OK! I've moved it to the script. We also pass a parameter to the `install.sh`, so it's more consistent indeed.

romintomasetti (approved):

PR #43, opened 2025-04-30
feat(callbacks): add timer listener

maartenarnst:
This PR adds a timer listener.

The purpose is to time a specific, possibly buried, kernel or region.


tests/callbacks/test_TimerListener.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include "kokkos-utils/callbacks/TimerListener.hpp"
+#include "kokkos-utils/timer/Duration.hpp"
+
+#include "tests/callbacks/TestWorkload.hpp"
+
+/**
+ * @addtogroup unittests
+ *
+ * @c Kokkos callback timer listener
+ * ------------------------------------
```
romintomasetti:
```suggestion
 * @c Kokkos callback timer listener
 * ---------------------------------
```

tests/callbacks/test_TimerListener.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+using execution_space = Kokkos::DefaultExecutionSpace;
+
+namespace Kokkos::utils::tests::callbacks
+{
+
+using namespace Kokkos::utils::callbacks;
+
+//! Listener to time parallel for regions.
+using parallel_for_timer_t = ParallelForTimerListener<EventRegexMatcher, execution_space>;
```
romintomasetti:
```suggestion
//! Listener to time parallel for regions whose label match a regex.
using parallel_for_timer_t = ParallelForTimerListener<EventRegexMatcher, execution_space>;
```

tests/callbacks/test_TimerListener.cpp
```diff
@@ -0,0 +1,109 @@
...
+    /// and @ref Kokkos::utils::callbacks::EndParallelForEvent.
+    static_assert(std::same_as<
+        listener_event_type_list_t<parallel_for_timer_t>,
+        Kokkos::Impl::type_list<BeginParallelForEvent, EndParallelForEvent>
+    >);
+
+    static_assert(Listener<parallel_for_timer_t>);
+}
+
+//! @test Check the behavior of @c Kokkos::utils::callbacks::ParallelForTimerListener.
```
romintomasetti:
```suggestion
//! @test Check the behavior of @ref Kokkos::utils::callbacks::ParallelForTimerListener.
```

tests/callbacks/test_TimerListener.cpp
```diff
@@ -0,0 +1,109 @@
...
+    ASSERT_GT(par_for_timer_matching->timer.duration(), Kokkos::utils::timer::seconds(0));
+
+    EXPECT_FALSE(par_for_timer_no_match->connected() || par_for_timer_no_match->closed())
+        << "The listener should not be connected or closed because the events did not match.";
+
+    Manager::unregister_listener(par_for_timer_matching.get());
+    Manager::unregister_listener(par_for_timer_no_match.get());
+}
+
+//! Listener to time profile regions.
```
romintomasetti:
```suggestion
//! Listener to time profile regions whose label match a regex.
```

tests/callbacks/test_TimerListener.cpp
```diff
@@ -0,0 +1,109 @@
...
+using region_timer_t = RegionTimerListener<EventRegexMatcher, execution_space>;
+
+//! @test Check traits of @c Kokkos::utils::callbacks::RegionTimerListener.
```
romintomasetti:
```suggestion
//! @test Check traits of @ref Kokkos::utils::callbacks::RegionTimerListener.
```

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+template <typename...>
+class TimerListener;
+
+/**
+ * @brief Listener that starts a timer when a begin event is received and stops it
+ *        when a corresponding end event is received.
+ *
+ * The start event is selected by requiring it to be of type @p BeginEventType and
+ * to be a match for the passed matcher. The stop event must be of type @p EndEventType
```
romintomasetti:
```suggestion
 * to be a match for @ref matcher. The stop event must be of type @p EndEventType
```

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,169 @@
...
+    Kokkos::utils::callbacks::Matcher<Kokkos::Impl::type_list<BeginEventType>> MatcherType,
+    Kokkos::utils::concepts::ExecutionSpace Exec
+>
+class TimerListener<BeginEventType, EndEventType, MatcherType, Exec>
+{
+public:
+    static constexpr uint64_t invalid_event_id = Kokkos::Experimental::finite_max_v<uint64_t>;
+
+public:
+    template <typename T> requires (std::same_as<std::remove_cvref_t<T>, MatcherType>)
```
romintomasetti: I guess it is not needed and in fact will constrain the constructor too much ? What if my matcher can be constructed from *e.g.* a `std::string` ?
```suggestion
    template <typename T>
```

tests/callbacks/test_TimerListener.cpp
```diff
@@ -0,0 +1,109 @@
...
+        Kokkos::Impl::type_list<PushRegionEvent, PopRegionEvent>
+    >);
+
+    static_assert(Listener<region_timer_t>);
+}
+
+//! @test Check the behavior of @c Kokkos::utils::callbacks::RegionTimerListener.
+TEST_F(TimerListenerTest, region)
+{
+    const auto par_for_timer = std::make_shared<parallel_for_timer_t>(exec, EventRegexMatcher(std::regex("computation - level 0 - pfor")));
```
romintomasetti: Using my suggestion to loosen the constructor constraint ?
```suggestion
    const auto par_for_timer = std::make_shared<parallel_for_timer_t>(exec, std::regex("computation - level 0 - pfor"));
```

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,169 @@
...
+    /**
+     * @todo Consider optimizing the implementation to avoid repeating the regex many
+     *       times after the timer is already closed.
+     *
+     * @todo Allow the possibility to combine multiple matchers with logical operators, so that
+     *       we can match e.g. a begin event with a specific name within a specific section.
+     */
+    void operator()(const BeginEventType& event) {
+        if (matcher(event)) {
+            if (this->connected() || this->closed()) Kokkos::abort("Timer cannot match a begin event twice.");
```
romintomasetti: I guess you can get a substantial reduction of the overhead of this timer by skipping any event that is not launched on `exec`. An integer comparison is much cheaper than a regex matching.

If you do it, please also:
- precompute the `dev_id` of `exec` once and forall ?
- add a test that launches *e.g.* two parallel reduce with the same label (so both would match) but one is on `exec` and the other one is on `exec_other`. You should only match one.

```suggestion
    void operator()(const BeginEventType& event)
    {
        if(event.dev_id != Kokkos::Tools::Experimental::device_id(exec)) return;

        if (matcher(event)) {
            if (this->connected() || this->closed()) Kokkos::abort("Timer cannot match a begin event twice.");
```

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,169 @@
...
+            if (this->connected() || this->closed()) Kokkos::abort("Timer cannot match a region twice.");
+
+            this->timer.reset(this->exec);
+            this->nested_level = 1;
+        } else {
+            if (this->connected()) ++this->nested_level;
+        }
+    }
+
+    void operator()(const Kokkos::utils::callbacks::PopRegionEvent& /* event */)
```
romintomasetti:
```suggestion
    void operator()(const Kokkos::utils::callbacks::PopRegionEvent&)
```

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,169 @@
...
+    /// When the timer has matched a pop event corresponding to each nested push event in the matched region,
+    /// if any, as well as to the matched push event, and the stack is thus empty, it is said to be "closed".
+    bool closed() const { return this->timer.is_valid(); }
+
+public:
+    Kokkos::utils::timer::Timer<Exec> timer {};
+
+private:
+    Exec exec {};
+    MatcherType matcher;
```
romintomasetti: This is non sense to use the timer for an event that has no exec. Use a regular host timer.

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: I tried but Doxygen gives an error saying that it doesn't know the reference `matcher`. I'm not sure what the issue is, but ll try more to make it work.

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: This is a good suggestion indeed, because there's quite some duplication and long lines of code now. We haven't done it for other similar cases. Probably best to do it only for the cases in this PR.

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: I changed it, but then I reverted some of the changes. 

The change that remains is that the `Exec` now defaults to `Kokkos::DefaultHostExecutionSpace` for the `RegionTimerListener`.

I reverted because I was thinking of cases where we'd want to time a group of several kernels, not just one, but that all go to a stream. Something like a graph that we'd submit to a stream. And so we may want to do a `PushRegion` before and after `PopRegion` after and record events in the stream.

Some food for thought left for what to do best :)

tests/callbacks/test_TimerListener.cpp
maartenarnst: OK! Works indeed!

tests/callbacks/test_TimerListener.cpp
maartenarnst: OK. Done!

tests/callbacks/test_TimerListener.cpp
maartenarnst: Done too!

tests/callbacks/test_TimerListener.cpp
maartenarnst: Indeed. Still a todo for me.

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: OK. Good suggestion indeed! I've modified it and added tests. The test of the case where we submit on one exec, but the TimerListener is on another exec didn't pass. The reason appears to be that we test only on `OpenMP`, and, on `OpenMP`, it appears that the `device_id` is the same for each instance:

https://github.com/kokkos/kokkos/blob/88485ca664e39a93725b44f99d3c4b03fcf701e7/core/src/OpenMP/Kokkos_OpenMP.hpp#L142

And so the `TimerListener` accepts the event, even if it's on another `exec`.

romintomasetti: Simply adapt the test by adding something like:
```c++
ASSERT_THAT(event accepted, only if exec A dev id equals exec B dev id);
```

It makes sense that the device ID is the same for several `OpenMP` instances, because they aren't memory pools and are on the same CPU (?).

include/kokkos-utils/callbacks/TimerListener.hpp
romintomasetti: OK. Probably good to add a note in the code.

include/kokkos-utils/callbacks/TimerListener.hpp
romintomasetti: And we should probably add a GPU runner.

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: OK. I've added two notes to the doc.

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: OK. Added such a test. Let's see what the new GPU runner does :)

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    //! When the listener has matched the begin, but not yet the end event, it is said to be "connected".
+    bool connected() const { return (this->event_id != invalid_event_id) && !this->closed(); }
+
+    //! When the listener has matched both the begin and end events, it is said to be "closed".
+    bool closed() const { return this->is_closed; }
+
+    //! Open the listener so that it can be reused to match a begin event again.
+    void open()
```
romintomasetti: This is new. Not sure to see:
- why we would reuse a matcher
- why you call this `open` and not `reset` ?

maartenarnst: It's new indeed! 

I added it to make things work when I used it in a benchmark with more than 1 iteration. The `SetUp` method of the benchmark creates the `TimerListener`. And then the benchmark loops over the `run` method, and so it iterates  through the begin/end, begin/end, ....

I used `open` instead of `reset` because we already use `reset` in the underlying `Timer`. And there, `reset` is not really "resetting" things, but rather recording the start event. Not sure what to do best. We could keep `open` here, which goes well with the `closed`. Another option would be to rename `reset` as `start` in `Timer` and then rename `open` here as `reset`.

romintomasetti: I agree with the last suggestion:
- rename `Timer::reset` to `Timer::start`
- rename `open` to `reset`

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,209 @@
...
+            if (this->connected() || this->closed()) Kokkos::abort("Listener cannot match a region twice.");
+
+            this->timer.reset(this->exec);
+            this->nested_level = 1;
+        } else {
+            if (this->connected()) ++this->nested_level;
+        }
+    }
+
+    void operator()(const Kokkos::utils::callbacks::PopRegionEvent&)
```
romintomasetti:
```suggestion
    void operator()(const PopRegionEvent&)
```

and similar occurrences in this file.

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,209 @@
...
+    explicit TimerListener(T&& matcher_)
+      : matcher(std::forward<T>(matcher_)) {}
+
+    template <typename T> requires (std::same_as<std::remove_cvref_t<T>, MatcherType>)
+    explicit TimerListener(Exec exec_, T&& matcher_)
+      : exec(std::move(exec_)), matcher(std::forward<T>(matcher_)) {}
+
+    void operator()(const Kokkos::utils::callbacks::PushRegionEvent& event)
+    {
+        if (matcher(event)) {
```
romintomasetti: Please try to implement this in terms of the `EventInRegionMatcher`.

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,209 @@
...
+ * @c Kokkos parallel for, parallel reduce, parallel scan, or deep copy operation, to which
+ * an execution space instance is passed. In such a situation, the execution space instance
+ * passed to this class's constructor should be the same as the one passed to the workload.
+ */
+template <
+    Kokkos::utils::callbacks::BeginEvent BeginEventType, Kokkos::utils::callbacks::EndEvent EndEventType,
+    Kokkos::utils::callbacks::Matcher<Kokkos::Impl::type_list<BeginEventType>> MatcherType,
+    Kokkos::utils::concepts::ExecutionSpace Exec
+>
+class TimerListener<BeginEventType, EndEventType, MatcherType, Exec>
```
romintomasetti: I think we should be smarter than that.

Why not having 2 matchers, one for the start, one for the stop ?

include/kokkos-utils/callbacks/EventBeginEndEventIdMatcher.hpp
```diff
@@ -21,7 +21,6 @@ struct EventBeginEndEventIdMatcher
     {
         if(matcher(event))
         {
-            if (event_id != invalid_event_id) Kokkos::abort("EventBeginEndEventIdMatcher cannot match a begin event twice.");
```
maartenarnst: ~~Let's consider renaming this file and class `EventBeginEndIdMatcher` to avoid repeating the `Event` twice in the name.~~

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+#include "kokkos-utils/callbacks/Listener.hpp"
+#include "kokkos-utils/callbacks/Matcher.hpp"
+#include "kokkos-utils/concepts/ExecutionSpace.hpp"
+
+namespace Kokkos::utils::callbacks
+{
+
+// template <typename...>
+// class TimerListener;
```
maartenarnst: Can now be deleted probably.

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+template <typename TimerType, typename EventType>
+concept TimerStartFor = Event<EventType> && requires (TimerType& timer, const EventType& event) {
+    { timer.start(event) } -> std::same_as<void>;
+};
```
maartenarnst: This concept could also be renamed `Startable` or `StartableWith`. Because it's just making sure that there's a `start` method, and it's not really checking we have a meaningful timer. Just `Startable` would be a bit like `std::invocable`.

include/kokkos-utils/callbacks/EventBeginEndEventIdMatcher.hpp
maartenarnst: Also, we could have the class store the types of the begin and end events.

maartenarnst: Or `BeginEndEventIdMatcher`.

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,113 @@
...
+template <typename TimerType, typename EventType>
+concept TimerStopFor = Event<EventType> && requires (TimerType& timer, const EventType& event) {
+    { timer.stop(event) } -> std::same_as<void>;
+};
+
+/**
+ * @brief Listener that starts a timer when a begin event is received and stops it
+ *        when a corresponding end event is received.
+ *
+ * The start event is selected by requiring it to be of type @p BeginEventType and
```
maartenarnst: If the `BeginEndMatcherType` stores the types of the begin and end evens, you can also use them to update this docstring here.

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,113 @@
...
+ * @c Kokkos parallel for, parallel reduce, parallel scan, or deep copy operation, to which
+ * an execution space instance is passed. In such a situation, the execution space instance
+ * passed to this class's constructor should be the same as the one passed to the workload.
+
+ TODO doc refactor
+ */
+template <
+    Matcher BeginEndMatcherType,
+    typename TimerType
+>
```
maartenarnst, line 44: If the `BeginEndMatcherType` stores the begin and end events, you can check here with a `requires` that the timer can be started and stopped with those event types. Those checks can then be deleted from the body of the function.

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,113 @@
...
+class TimerListener
+{
+public:
+    // check if needed
+    template <typename T, typename U = TimerType>
+    TimerListener(T&& matcher_, U&& timer_ = TimerType{})
+      : matcher(std::forward<T>(matcher_)), timer(std::forward<U>(timer_)) {printf("TimerListener: constructor\n");}
```
maartenarnst: There's a check and a print left 🧹 😄

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,113 @@
...
+     * @todo Consider optimizing the implementation to avoid repeating the regex many
+     *       times after the timer is already closed.
+     *
+     * @todo Allow the possibility to combine multiple matchers with logical operators, so that
+     *       we can match e.g. a begin event with a specific name within a specific section.
+     * Since we evaluate possibly both begin and end, we could check that it's possible to do a start stop in one call.
+     * In one call we can let the end matcher get the paralle for event id for instance.
+     */
+    template <Event EventType> requires MatcherFor<BeginEndMatcherType, EventType>
+    void operator()(const EventType& event)
```
maartenarnst, line 48: If the `BeginEndMatcherType` stores the begin and end types, we can split this function in two functions.

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,113 @@
...
+    {
+        const bool matching = matcher(event);
+
+        // it matches and it was not start yet
```
maartenarnst:
```suggestion
        // The begin event matches and the timer has not been started yet
```

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,113 @@
...
+        if(!this->matched_begin && matching)
+        {
+            if constexpr (TimerStartFor<TimerType, EventType>)
+            {
+                this->matched_begin = true;
+                this->timer.start(event);
+            }
+        }
+        // and when it's started and stops matching stop timer
```
maartenarnst:
```suggestion
        // The end event matches and the timer was started but has not been stopped yet.
```

include/kokkos-utils/callbacks/TimerListenerBeginEnd.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+namespace Kokkos::utils::callbacks
+{
+
+//! Timer for events that are enqueued on @ref exec.
+template <Kokkos::utils::concepts::ExecutionSpace Exec>
+struct EnqueuedEventTimer
+{
+    //! Start the timer. The event must be on @ref exec.
+    template <OnDeviceEvent EventType>
```
maartenarnst: We could also rename this concept `DeviceIdentifiableEvent`.

include/kokkos-utils/callbacks/TimerListenerBeginEnd.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+    Exec exec;
+    uint32_t dev_id = Kokkos::Tools::Experimental::device_id(exec);
+    Kokkos::utils::timer::Timer<Exec> timer {};
+};
+
+template <Matcher MatcherType, BeginEvent BeginEventType, EndEvent EndEventType, Kokkos::utils::concepts::ExecutionSpace Exec> requires MatcherFor<MatcherType, BeginEventType>
+struct BeginEndTimerListener : public TimerListener<EventBeginEndEventIdMatcher<EventOnExecMatcher<MatcherType, Exec>, BeginEventType, EndEventType>, EnqueuedEventTimer<Exec>>
+{
+    using enqueued_t = EventOnExecMatcher<MatcherType, Exec>;
```
maartenarnst: We could rename `EventOnExecMatcher` as `EventDeviceIdMatcher`.

maartenarnst: Not sure where to put it, but `EventRegionMatcher` could become `EventInRegionMatcher` to be consistent with `EventInProfileSection`.

include/kokkos-utils/callbacks/TimerListenerBeginEnd.hpp
```diff
@@ -0,0 +1,74 @@
...
+    using matcher_t  = EventBeginEndEventIdMatcher<enqueued_t, BeginEventType, EndEventType>;
+    using timer_t    = EnqueuedEventTimer<Exec>;
+    using base_t     = TimerListener<matcher_t, timer_t>;
+
+    template <typename T>
+    BeginEndTimerListener(T&& m, const Exec& exec) : base_t(
+        matcher_t{.matcher = enqueued_t{.matcher = std::forward<T>(m), .exec = exec}},
+        timer_t{.exec = exec}
+    ) {printf("BeginEndTimerListener constructor\n");}
```
maartenarnst: We can probably delete the printout.

include/kokkos-utils/callbacks/TimerListenerRegion.hpp
```diff
@@ -0,0 +1,58 @@
+#ifndef KOKKOS_UTILS_CALLBACKS_TIMERLISTENERREGION_HPP
+#define KOKKOS_UTILS_CALLBACKS_TIMERLISTENERREGION_HPP
```
maartenarnst: The file should probably be `RegionTimerListener.hpp`.

include/kokkos-utils/callbacks/TimerListenerBeginEnd.hpp
maartenarnst: If you make this change, the timer can also be renamed `DeviceIdentifiableEventTimer`.

include/kokkos-utils/timer/Timer.hpp
```diff
@@ -50,6 +53,35 @@ class Timer
     bool    started = false, stopped = false;
 };
 
+template <>
+class Timer<void>
```
maartenarnst: Ultimately, this timer is the same as our current Timer<Exec> for host execution spaces, right? We could try to avoid the repetition. We could see if we could define this `Timer<void>` first. Then, we define `Timer<Exec>`, and we say that the instantiations for host execution spaces are equal to the `Timer<void>`.

tests/callbacks/test_TimerListener.cpp
```diff
@@ -0,0 +1,272 @@
...
+    ASSERT_GT(par_for_timer->timer.duration(), Kokkos::utils::timer::milliseconds{0.});
+    ASSERT_LT(par_for_timer->timer.duration(), region_timer->timer.duration());
+
+    Manager::unregister_listener(par_for_timer.get());
+    Manager::unregister_listener(region_timer.get());
+}
+
+//! Mesure the time for launch + time for kernel.
+template <Kokkos::utils::concepts::ExecutionSpace Exec>
+struct EnqueuedEventTimerWithLatency
```
maartenarnst: OK. This is nice indeed! We could also move it to its own PR. But it's fine to keep it here too.

maartenarnst:
OK! Voila voila! Two main suggestions:
- We could let `BeginEndEventIdMatcher` store the types of the begin and end events. And then simplify some code because we now know the types.
- We could see if we can avoid the duplication (?) between `Timer<void>` and `Timer<Exec>` for host execution spaces.

And a few suggestions for renaming. Sorry :)

include/kokkos-utils/callbacks/EventBeginEndEventIdMatcher.hpp
romintomasetti: I've tried to start with *Event* for matchers related to events. So I'm not in favor of `BeginEndEventIdMatcher`, but I agree it's weird repeating.

How about we just do `EventBeginEndIdMatcher`?

include/kokkos-utils/callbacks/TimerListener.hpp
romintomasetti: Sure, done.

include/kokkos-utils/callbacks/TimerListener.hpp
romintomasetti: It's indeed better, but I think that `Startable` still misses the *Event* related part.

How about `StartableWithEvent` ? It would look like:
```c++
template <StartableWithEvent<BeginFenceEvent> TimerType>
...
```

include/kokkos-utils/callbacks/TimerListener.hpp
romintomasetti: I'll update the docstrings once the code is converged (remember that I'm inheriting your old documentation too).

include/kokkos-utils/callbacks/TimerListener.hpp
romintomasetti: The current design wouldn't allow such a thing.

For instance, it's allowed you do:
```c++
using my_timer_t = TimerListener<EventRegexMatcher, TimerType>;
```
and `EventRegexMatcher` hasn't any "begin" and "end" event type.

include/kokkos-utils/callbacks/TimerListener.hpp
romintomasetti: We need to discuss if we want `TimerListener` to continue make its `matcher` private, in which case we need to provide this constructor.

include/kokkos-utils/callbacks/TimerListener.hpp
romintomasetti: Related to my comment above, not possible with the current design.

include/kokkos-utils/callbacks/TimerListenerBeginEnd.hpp
romintomasetti: I guess it would be better to call it `EnqueuedEvent` ?

include/kokkos-utils/callbacks/TimerListenerBeginEnd.hpp
romintomasetti: See:
- #67

include/kokkos-utils/callbacks/TimerListenerBeginEnd.hpp
romintomasetti: :heavy_check_mark:

include/kokkos-utils/timer/Timer.hpp
romintomasetti: With `void`, both `start` and `stop` don't take an `exec` argument.

include/kokkos-utils/callbacks/TimerListenerBeginEnd.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include "kokkos-utils/callbacks/TimerListener.hpp"
+#include "kokkos-utils/timer/Duration.hpp"
+#include "kokkos-utils/timer/Timer.hpp"
+
+namespace Kokkos::utils::callbacks
+{
+
+//! Timer for events that are enqueued on @ref exec.
+template <Kokkos::utils::concepts::ExecutionSpace Exec>
+struct EnqueuedEventTimer
```
maartenarnst: We can probably delete the enqueuedeventimer because it's covered by the other matcher.

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,110 @@
...
+     *       times after the timer is already closed.
+     *
+     * @todo Allow the possibility to combine multiple matchers with logical operators, so that
+     *       we can match e.g. a begin event with a specific name within a specific section.
+     * Since we evaluate possibly both begin and end, we could check that it's possible to do a start stop in one call.
+     * In one call we can let the end matcher get the paralle for event id for instance.
+     */
+    template <Event EventType> requires MatcherFor<BeginEndMatcherType, EventType>
+    void operator()(const EventType& event)
+    {
```
maartenarnst, line 49: Let's add a "return" if we're closed already.

include/kokkos-utils/callbacks/TimerListenerBeginEnd.hpp
romintomasetti: Let's try.

include/kokkos-utils/callbacks/TimerListener.hpp
romintomasetti: OK.

include/kokkos-utils/callbacks/TimerListenerBeginEnd.hpp
romintomasetti: It's actually not possible, because neither `TimerListener` nor our `Kokkos::utils::timer::Timer` store the `exec` instance.

include/kokkos-utils/callbacks/EventBeginEndEventIdMatcher.hpp
maartenarnst: Indeed, we can do `EventBeginEndIdMatcher`.

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: I would just go with `StartableWith` in that case. A bit like `std::constructible_from`.

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,112 @@
+#ifndef KOKKOS_UTILS_CALLBACKS__TIMERLISTENER_HPP
+#define KOKKOS_UTILS_CALLBACKS__TIMERLISTENER_HPP
```
maartenarnst: No need for the two dashes.

romintomasetti: :heavy_check_mark:

include/kokkos-utils/callbacks/TimerListener.hpp
romintomasetti: :heavy_check_mark:

include/kokkos-utils/callbacks/Matcher.hpp
```diff
@@ -1,6 +1,8 @@
 #ifndef KOKKOS_UTILS_CALLBACKS_MATCHER_HPP
 #define KOKKOS_UTILS_CALLBACKS_MATCHER_HPP
 
+#include <concepts>
```
maartenarnst: Looks like a leftover. I'll delete.

include/kokkos-utils/callbacks/TimerListenerBeginEnd.hpp
maartenarnst: OK. Made it `EnqueuedEvent`.

include/kokkos-utils/callbacks/TimerListenerBeginEnd.hpp
maartenarnst: We kept it. Because it paves the way to become a bit more general later.

include/kokkos-utils/callbacks/TimerListenerBeginEnd.hpp
maartenarnst: OK. Got it.

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: Outdated.

tests/callbacks/test_TimerListener.cpp
maartenarnst: OK! Works now indeed! Good suggestion!

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: Solved indeed!

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: OK. Done indeed as said in an earlier thread. Done everywhere.

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: Done too!

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: Still there 😄

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: Outdated.

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: OK. Done now thanks to the new moduler design of the `TimerListener`. Good suggestion indeed!

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: In our new design we should be able to support such use cases, because we now have a very general `BeginEndMatcherType` template argument. But our current specialisations `BeginEndTimerListener` and `RegionListener` still select based on a match for the "begin event" and then look for a "paired" "end event".

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: Indeed. Got it!

include/kokkos-utils/timer/Timer.hpp
maartenarnst: Done as part of:
- #74

tests/callbacks/test_TimerListener.cpp
maartenarnst: Still in this PR for now.

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: We went with `StartableWithoutEvent` and `StartableWithEvent`.

include/kokkos-utils/callbacks/EventQueueMatcher.hpp
```diff
@@ -8,19 +8,14 @@ namespace Kokkos::utils::callbacks
 {
 
 //! Match an event whose @c dev_id is the same as the one of @ref exec.
-template <Matcher MatcherType, Kokkos::utils::concepts::ExecutionSpace Exec>
+template <Kokkos::utils::concepts::ExecutionSpace Exec>
```
romintomasetti: Indeed, let's try to use some `And<MatcherType, EventQueueMatcher>` instead.

tests/callbacks/test_Matchers.cpp
```diff
@@ -225,7 +225,7 @@ TEST(EventBeginEndIdMatcher, traits)
 //! @test Check the behavior of @ref Kokkos::utils::callbacks::EventBeginEndIdMatcher.
 TEST(EventBeginEndIdMatcher, operator_parentheses)
 {
-    EventBeginEndIdMatcher<EventNameMatcher, BeginParallelForEvent> matcher {.matcher = EventNameMatcher("example-pfor")};
+    EventBeginEndIdMatcher<EventNameMatcher, BeginParallelForEvent> matcher {.matcher = {"example-pfor"}};
```
romintomasetti: Let's extract these changes to a separate PR :wink:

maartenarnst: Indeed! It's:
- https://github.com/uliegecsm/kokkos-utils/pull/80

include/kokkos-utils/callbacks/EventQueueMatcher.hpp
maartenarnst: Done. Thanks to:
- https://github.com/uliegecsm/kokkos-utils/pull/78

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+#include "kokkos-utils/callbacks/Listener.hpp"
+#include "kokkos-utils/callbacks/Matcher.hpp"
+#include "kokkos-utils/concepts/ExecutionSpace.hpp"
+
+namespace Kokkos::utils::callbacks
+{
+
+template <typename TimerType>
+concept StartableWithoutEvent = requires (TimerType& timer) {
```
romintomasetti: We could name the "no argument" version simply as `Startable` and `Stoppable`.

tests/callbacks/test_TimerListener.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+    static_assert(std::constructible_from<
+        region_timer_with_enqueued_event_timer_t,
+        std::regex, execution_space
+    >);
+}
+
+//! Timer that measures the time to launch, in addition to the time to execute the kernel.
+template <Kokkos::utils::concepts::ExecutionSpace Exec>
+struct EnqueuedEventTimerWithLatency
```
romintomasetti: Let's move this one to its own include file :wink: (To make it visible)

tests/callbacks/test_TimerListener.cpp
```diff
@@ -0,0 +1,288 @@
...
+    const auto par_for_timer = std::make_shared<parallel_for_timer_with_latency_t>(
+        "computation - level 0 - pfor", exec
+    );
+
+    Manager::register_listener(par_for_timer);
+
+    MyWorkload<execution_space>{}.execute(exec);
+
+    ASSERT_GT(par_for_timer->timer.duration(), Kokkos::utils::timer::milliseconds{0.});
+    ASSERT_GT(par_for_timer->timer.latency(),  Kokkos::utils::timer::milliseconds{0.});
```
romintomasetti: Let's add a check that latency was greater than 0 but smaller than the time for the call to `MyWorkload`.

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: OK. Better indeed!

tests/callbacks/test_TimerListener.cpp
maartenarnst: OK. Done. It's PR:
- #81

tests/callbacks/test_TimerListener.cpp
maartenarnst: We'll need to do it in:
- #81

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,109 @@
+#ifndef KOKKOS_UTILS_CALLBACKS_TIMERLISTENER_HPP
+#define KOKKOS_UTILS_CALLBACKS_TIMERLISTENER_HPP
+
+#include "Kokkos_NumericTraits.hpp"
```
romintomasetti: Probably a leftover.

romintomasetti:
```suggestion
```

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+#include "kokkos-utils/callbacks/Listener.hpp"
+#include "kokkos-utils/callbacks/Matcher.hpp"
+#include "kokkos-utils/concepts/ExecutionSpace.hpp"
```
romintomasetti: Probably a leftover.

```suggestion
```

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,109 @@
...
+ * @ref reset, a new begin event can be matched.
+ */
+template <
+    Matcher BeginEndMatcherType,
+    typename TimerType
+>
+class TimerListener
+{
+public:
+    //! @todo Check if this constructor is needed.
```
romintomasetti: Have you checked this ?

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,109 @@
...
+      : matcher{std::forward<T>(matcher_)}, timer{std::forward<U>(timer_)} {}
+
+    template <Event EventType> requires MatcherFor<BeginEndMatcherType, EventType>
+    void operator()(const EventType& event)
+    {
+        if(closed()) return;
+
+        const bool matching = matcher(event);
+
+        // Start when the event matches while no begin event has been matched before.
```
romintomasetti:
```suggestion
        //! Start when the event matches while no begin event has been matched before.
```

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,109 @@
...
+        {
+            this->matched_begin = true;
+
+            if constexpr (Startable<TimerType>) {
+                this->timer.start();
+            } else if constexpr (StartableWithEvent<TimerType, EventType>) {
+                this->timer.start(event);
+            }
+        }
+        // Stop when the event matches after a begin event has been matched.
```
romintomasetti:
```suggestion
        //! Stop when the event matches after a begin event has been matched.
```

include/kokkos-utils/callbacks/TimerListener.hpp
```diff
@@ -0,0 +1,109 @@
...
+        else if(this->matched_begin && ! this->matched_end && matching)
+        {
+            if constexpr (Stoppable<TimerType>) {
+                this->timer.stop();
+            } else if constexpr (StoppableWithEvent<TimerType, EventType>) {
+                this->timer.stop(event);
+            }
+
+            this->matched_end = true;
```
romintomasetti: For the symmetry with the begin case.
```suggestion
            this->matched_end = true;
   
            if constexpr (Stoppable<TimerType>) {
                this->timer.stop();
            } else if constexpr (StoppableWithEvent<TimerType, EventType>) {
                this->timer.stop(event);
            }
```

include/kokkos-utils/callbacks/RegionTimerListener.hpp
```diff
@@ -0,0 +1,43 @@
+#ifndef KOKKOS_UTILS_CALLBACKS_TIMERLISTENERREGION_HPP
```
romintomasetti:
```suggestion
#ifndef KOKKOS_UTILS_CALLBACKS_REGIONTIMERLISTENER_HPP
```

include/kokkos-utils/callbacks/BeginEndTimerListener.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include "kokkos-utils/callbacks/EventBeginEndIdMatcher.hpp"
+#include "kokkos-utils/callbacks/EventQueueMatcher.hpp"
+#include "kokkos-utils/callbacks/TimerListener.hpp"
+
+namespace Kokkos::utils::callbacks
+{
+
+/**
+ * @brief Specialization of @ref TimerListener that is well-suited for timing the execution of a
+ *        parallel workload to which an execution space instance is passed.
```
romintomasetti:
```suggestion
 * @brief Timer listener that is well-suited for timing the execution of a
 *        parallel workload enqueued on a particular execution space instance.
```

include/kokkos-utils/callbacks/BeginEndTimerListener.hpp
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * This specialization is particularly well-suited for timing a @c Kokkos parallel for, parallel
+ * reduce, parallel scan, or deep copy operation,  In such a situation, the execution space
```
romintomasetti:
```suggestion
 * reduce, parallel scan, or deep copy operation.  In such a situation, the execution space
```

include/kokkos-utils/callbacks/BeginEndTimerListener.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * instance passed to this class's constructor should be the same as the one passed to the workload.
+ *
+ * This class uses as timer a @ref EnqueuedEventTimer, instantiated for the execution space @p Exec.
+ * This timer stores the execution space instance passed to the constructor, and uses it as the one
+ * passed to the @c start and @c stop member functions of the underlying @ref Kokkos::utils::timer::Timer.
+ */
```
romintomasetti: No need to describe here how `EnqueuedEventTimer` works.

tests/callbacks/test_TimerListener.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+#include "tests/callbacks/TestWorkload.hpp"
+
+/**
+ * @addtogroup unittests
+ *
+ * @c Kokkos callback timer listener
+ * ---------------------------------
+ *
+ * This group of tests check the behavior of @ref Kokkos::utils::callbacks::TimerListener.
```
romintomasetti:
```suggestion
 * This group of tests check the behavior of @ref Kokkos::utils::callbacks::TimerListener and related.
```

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: I've just tried to check it. It seems we currently need it, and it seems the issue is that
- the `matcher` member is private
- the order of the members is such that `matched_begin` and `matched_begin` "are in the middle"

If I make all the members public and begin with `matcher` and `timer`, the constructor can be deleted.

It seems the issue of private members comes up in other classes as well. So probably best discussed live if we want to make all the members public.

include/kokkos-utils/callbacks/TimerListener.hpp
maartenarnst: Indeed. Good catch! I've deleted this include here.

It seems a related issue is that we're not very consistent yet with "include what you need". So probably best discussed live what we'd want to do.

tests/callbacks/test_TimerListener.cpp
```diff
@@ -0,0 +1,212 @@
...
+        std::regex("computation - level 0 - pfor"), exec
+    );
+    const auto par_for_timer_no_match = std::make_shared<parallel_for_timer_t>(
+        std::regex("this-should-not-match"), exec
+    );
+
+    Manager::register_listener(par_for_timer_matching);
+    Manager::register_listener(par_for_timer_no_match);
+
+    // TODO check which callbacks are activated at this point
```
romintomasetti: Leftover ?

maartenarnst: It's probably a test we may want to add to the manager tests.

romintomasetti (approved):

PR #45, opened 2025-05-04
ci(cuda): enable testing on blackwell120 runner

romintomasetti:
I need `Kokkos` *4.6.01* to use `BACKWELL120` (https://github.com/kokkos/kokkos/commit/51b7c713dd636bbd6bb31d624efec7c505f29a55).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,7 +45,7 @@ jobs:
                       compiler_family: gcc
                       base_image     : nvidia/cuda:12.8.1-devel-ubuntu24.04
                       platforms      : linux/amd64
-                      kokkos_arch    : VOLTA70
+                      kokkos_arch    : BLACKWELL120
                     - backend        : HIP
                       compiler_family: rocm
                       base_image     : rocm/dev-ubuntu-24.04:6.4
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -89,7 +89,7 @@ jobs:
               with:
                   context: .
                   platforms: ${{ matrix.platforms }}
-                  push: ${{ github.ref == 'refs/heads/develop' }}
+                  push: true #${{ github.ref == 'refs/heads/develop' }}
```
romintomasetti: Don't forget to revert this before merging.

maartenarnst:
OK! The compile flags look like:

```
tests/impl/CMakeFiles/test_type_traits.dir/flags.make:CXX_FLAGS = -O3 -DNDEBUG -std=c++20 -Wall -Wextra -Werror --Werror=cross-execution-space-call,all-warnings -extended-lambda -Wext-lambda-captures-this -expt-relaxed-constexpr -arch=sm_120 -fopenmp
```

So it seems the flags are correctly passed from `kokkos` to `kokkos-utils` and work.

.github/workflows/build.yml
maartenarnst: Done :)

maartenarnst (approved):
All good!

PR #48, opened 2025-05-05
callbacks: add "in region" matcher

romintomasetti:
## Summary

This is somewhat based on the `EventInProfileSectionRegexMatcher`, but in this case I've templated my matcher to make it more "composable".

## Related

- can probably be reused in #43

### DIFF
--- a/include/kokkos-utils/callbacks/EventInRegionMatcher.hpp
+++ b/include/kokkos-utils/callbacks/EventInRegionMatcher.hpp
@@ -1,0 +1,66 @@
+#ifndef KOKKOS_UTILS_CALLBACKS_EVENTINREGIONMATCHER_HPP
+#define KOKKOS_UTILS_CALLBACKS_EVENTINREGIONMATCHER_HPP
+
+#include "kokkos-utils/callbacks/Events.hpp"
+#include "kokkos-utils/callbacks/Matcher.hpp"
+
+namespace Kokkos::utils::callbacks
+{
+
+/**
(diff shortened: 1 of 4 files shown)


include/kokkos-utils/callbacks/EventInRegionMatcher.hpp
```diff
@@ -0,0 +1,62 @@
...
+ */
+template <typename MatcherType> requires Matcher<MatcherType, Kokkos::Impl::type_list<PushRegionEvent>>
+struct EventInRegionMatcher
+{
+    static constexpr uint32_t invalid_nested_level = Kokkos::Experimental::finite_max_v<uint32_t>;
+
+    //! If the @p event matches @ref matcher, @ref matching is set to @c true.
+    bool operator()(const PushRegionEvent& event)
+    {
+        if (this->matching) {
```
maartenarnst: We could move the `{` to the next line to keep our way of formatting "multiline scopes". There are a few other cases in the next lines of code too.

maartenarnst (dismissed):
OK. Good contribution indeed! And makes sense to do the follow up for the profile section as well. 🟢

include/kokkos-utils/callbacks/EventInRegionMatcher.hpp
romintomasetti: Indeed, and done :heavy_check_mark:

romintomasetti:
I'll merge since you approved and I only did formatting changes you requested.

PR #49, opened 2025-05-05
callbacks(profile section): make matcher more flexible

romintomasetti:
## Summary

Adds more flexibility to `EventInProfileSection[~~Regex~~]Matcher` :arrow_right: `EventInProfileSectionMatcher`.

This would *e.g.* enable someone to compose several matchers to tell if the profile section matches.

### DIFF
--- a/tests/callbacks/test_RecorderListener.cpp
+++ b/tests/callbacks/test_RecorderListener.cpp
@@ -5,7 +5,7 @@
 
 #include "kokkos-utils/impl/type_traits.hpp"
 
-#include "kokkos-utils/callbacks/EventInProfileSectionRegexMatcher.hpp"
+#include "kokkos-utils/callbacks/EventInProfileSectionMatcher.hpp"
 #include "kokkos-utils/callbacks/EventRegexMatcher.hpp"
 #include "kokkos-utils/callbacks/Helpers.hpp"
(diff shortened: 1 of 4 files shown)


romintomasetti:
@maartenarnst If you have 10 seconds to have a look :eyes: and approve :heavy_check_mark:

tests/callbacks/test_RecorderListener.cpp
```diff
@@ -78,7 +78,7 @@ TYPED_TEST(RecorderListenerSingleEventTypeTest, record)
 //! @test Check the behavior of @ref Kokkos::utils::callbacks::RecorderListener.
 TEST_F(RecorderListenerTest, recorded_events)
 {
-    EventInProfileSectionRegexMatcher matcher(std::regex("profile section"));
+    EventInProfileSectionMatcher matcher{.matcher = EventRegexMatcher{.regex = std::regex("profile section")}};
```
maartenarnst, line 81: OK! That's interesting. Didn't know that would work without specifying the template parameter. Looks like it's a C++20 feature.

maartenarnst (approved):
🟢 !

tests/callbacks/test_RecorderListener.cpp
romintomasetti: It's basic CTAD, since C++ 17.

PR #50, opened 2025-05-06
callbacks: add a conjunction matcher

romintomasetti:
## Summary

This PR introduces a `ConjunctionMatcher` that match **only if** all of its "internal" matchers match.

### DIFF
--- a/include/kokkos-utils/callbacks/ConjunctionMatcher.hpp
+++ b/include/kokkos-utils/callbacks/ConjunctionMatcher.hpp
@@ -1,0 +1,37 @@
+#ifndef KOKKOS_UTILS_CALLBACKS_CONJUNCTIONMATCHER_HPP
+#define KOKKOS_UTILS_CALLBACKS_CONJUNCTIONMATCHER_HPP
+
+#include "kokkos-utils/callbacks/Events.hpp"
+#include "kokkos-utils/callbacks/Matcher.hpp"
+
+namespace Kokkos::utils::callbacks
+{
+
+/**
+ * @brief Conjunction of matchers that is @c true only if all @ref matchers agree.
+ *
+ * @warning It is short-circuiting.
+ */
+template <Matcher... MatcherTypes> requires (sizeof...(MatcherTypes) > 1)
+struct ConjunctionMatcher
+{
+    template <typename... Args> requires (sizeof...(Args) == sizeof...(MatcherTypes))
+    explicit ConjunctionMatcher(Args&&... args) : matchers(std::forward<Args>(args)...) {}
+
+    template <Event EventType> requires (MatcherFor<MatcherTypes, EventType> && ...)
+    bool operator()(const EventType& event)
+    {
+        return std::apply([&event] (MatcherTypes&... matchers) {
+                return (matchers(event) && ...);
+        }, matchers);
+    }
+
+    std::tuple<MatcherTypes...> matchers;
+};
+
+template <typename... MatcherTypes>
+ConjunctionMatcher(MatcherTypes&&...) -> ConjunctionMatcher<MatcherTypes...>;
+
+} // namespace Kokkos::utils::callbacks
+
+#endif // KOKKOS_UTILS_CALLBACKS_CONJUNCTIONMATCHER_HPP
--- a/tests/callbacks/test_Matchers.cpp
+++ b/tests/callbacks/test_Matchers.cpp
@@ -1,9 +1,11 @@
+#include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
 #include "Kokkos_Core.hpp"
 
 #include "kokkos-utils/impl/type_traits.hpp"
 
+#include "kokkos-utils/callbacks/ConjunctionMatcher.hpp"
 #include "kokkos-utils/callbacks/EventBeginEndEventIdMatcher.hpp"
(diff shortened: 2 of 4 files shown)


include/kokkos-utils/callbacks/ConjunctionMatcher.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include "kokkos-utils/callbacks/Events.hpp"
+
+namespace Kokkos::utils::callbacks
+{
+/**
+ * @brief Conjunction of matchers that is @c true only if all @ref matchers agree.
+ *
+ * @todo Try to improve usability. For now, the construction requires a @c std::tuple,
+ *       and it would not be an aggregate if we add a user-defined constructor.
```
maartenarnst: Perhaps we should accept that it's not an aggregate and write a constructor?

An alternative may be to try something like 

```
template <typename... Matchers>
struct ConjunctionMatcher : public Matchers...
{
    ...
};
```
but we'd have to deal with the issue that the `Matchers` can all be of the same type. So the multiple inheritance would have to introduce an index or something to distinguish the base classes. May become too complex.

include/kokkos-utils/callbacks/ConjunctionMatcher.hpp
```diff
@@ -0,0 +1,30 @@
...
+ */
+template <typename... Matchers>
+struct ConjunctionMatcher
+{
+    template <Event EventType>
+    bool operator()(const EventType& event)
```
maartenarnst: Just wondering if we should write a requires clause to constrain the EventType based on the event types that the child matchers can match? Probably good to write a test where one of the child matchers matches only a subset of the events, and then use the `Matcher` concept to see whether that subset gets propagated to the `ConjunctionMatcher`.

tests/callbacks/test_Matchers.cpp
```diff
@@ -136,4 +137,42 @@ TEST(AnyEventMatcher, operator_parentheses)
...
+
+    ASSERT_FALSE(matcher(PushRegionEvent{.name = "not-the-one-we-want"}));
+
+    ASSERT_FALSE(matcher(AllocateDataEvent{}));
+
+    ASSERT_FALSE(matcher(PushRegionEvent{.name = "buried-region-one"}));
+
+    ASSERT_FALSE(matcher(AllocateDataEvent{}));
+
+    ASSERT_FALSE(matcher(PushRegionEvent{.name = "buried-region-two"}));
```
maartenarnst, line 324: OK! It's a nice use case indeed!

include/kokkos-utils/callbacks/ConjunctionMatcher.hpp
romintomasetti: Yep, much too much complex for such a basic functionality!

romintomasetti: Using a constructor and a deduction guide was enough !

include/kokkos-utils/callbacks/ConjunctionMatcher.hpp
romintomasetti: It could be done, but I guess it's too heavy for the compiler.

romintomasetti: OK, it's done !

It was indeed necessary to constrain the call operator of `ConjunctionMatcher` to only accept events that are accepted by all sub-matchers, otherwise it was reporting as a matcher for *any event but would fail compiling.

romintomasetti:
@maartenarnst I spent the day improving stuff, now it should be very clean !

include/kokkos-utils/callbacks/ConjunctionMatcher.hpp
```diff
@@ -0,0 +1,32 @@
...
+{
+    template <typename... Args> requires (sizeof...(Args) == sizeof...(Matchers))
+    explicit ConjunctionMatcher(Args&&... args) : matchers(std::forward<Args>(args)...) {}
+
+    template <Event EventType> requires (MatcherFor<Matchers, EventType> && ...)
+    bool operator()(const EventType& event)
+    {
+        return std::apply([&event] (Matchers&... matchers) {
+                return (matchers(event) && ...);
+        }, matchers);
```
romintomasetti: We could avoid evaluating the second sub-matcher if the first one is already `false`.

romintomasetti: Because `std::conjunction` is short-circuiting, so should this one.

And we'd have another helper called `AndMatcher` that would evaluate all sub-matchers and perform `AND`.

@maartenarnst What do you think ?

romintomasetti: It seems it's already all good [as stated in this order of evaluation rule](https://en.cppreference.com/w/cpp/language/eval_order):
> 6) The first (left) operand of the built-in [logical](https://en.cppreference.com/w/cpp/language/operator_logical) AND operator &&, the built-in logical OR operator || and the built-in [comma operator](https://en.cppreference.com/w/cpp/language/operator_other#Built-in_comma_operator) , is sequenced before the second (right) operand.

romintomasetti (commented):
Fine with me.

maartenarnst:
OK. Let's merge! Will be useful indeed.

maartenarnst (approved):
🟢

PR #51, opened 2025-05-07
callbacks(listener): refactor concept to make it faster

romintomasetti:
## Summary

This PR refactors the `Listener` concept to make it (much) faster.

## Description

Before this PR, the `Listener` concept would first collect all the event types that a callable can be a listener for and then check that the number of collected events is larger than 0.

Now, it uses `std::disjunction` which is short-circuiting.

IMO it is also more readable now.

To be noted that I hit a *internal linkage* issue with `std::disjunction` with **nvcc**. I sidestepped the issue with a preporcessing *if/else*.

## Related

- needs #53

### DIFF
--- a/include/kokkos-utils/callbacks/Listener.hpp
+++ b/include/kokkos-utils/callbacks/Listener.hpp
@@ -9,7 +9,11 @@ namespace Kokkos::utils::callbacks
 namespace impl
 {
 
-//! Helper struct needed for the implementation of @ref Kokkos::utils::callbacks::listener_event_type_list_t.
+/**
+ * Helper struct needed for the implementation of concepts and type traits such as:
+ *  - @ref Kokkos::utils::callbacks::listener_event_type_list_t
+ *  - @ref Kokkos::utils::callbacks::ListenerFor
+ */
 template <typename Callable>
 struct IsListenerFor
 {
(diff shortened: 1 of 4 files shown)


include/kokkos-utils/callbacks/Listener.hpp
```diff
@@ -19,16 +23,20 @@ struct IsListenerFor
...
- *        have a non-empty list of event types that it can be a listener for.
+ * @brief @p Callable is a listener if it is invocable with at least one event type
+ *        from @ref Kokkos::utils::callbacks::EventTypeList passed by @c const reference and returns @c void.
  */
 template <typename Callable>
-concept Listener = ( ! std::same_as<listener_event_type_list_t<Callable>, Kokkos::Impl::type_list<>>);
+concept Listener = Kokkos::utils::impl::type_list_any_v<impl::IsListenerFor<Callable>::template type, EventTypeList>;
+
+//! Check that @p Callable is a listener for each event in @p EventTypes.
+template <typename Callable, typename EventTypes>
```
romintomasetti:
```suggestion
template <typename Callable, typename... EventTypes>
```

include/kokkos-utils/impl/type_list.hpp
```diff
@@ -157,9 +157,20 @@ inline constexpr bool type_list_all_v = type_list_all<UnaryPred, TypeList>::valu
 template <template <typename> class UnaryPred, class List>
 struct type_list_any;
 
+/// As of @c Cuda 12.8, using @c std::conjunction sometimes ends up erroring out with
```
romintomasetti:
```suggestion
/// As of @c Cuda 12.8, using @c std::disjunction sometimes ends up erroring out with
```

PR #54, opened 2025-05-07
ci(cuda): update to `12.9.0`

romintomasetti:


.github/workflows/build.yml
```diff
@@ -89,7 +89,7 @@ jobs:
               with:
                   context: .
                   platforms: ${{ matrix.platforms }}
-                  push: ${{ github.ref == 'refs/heads/develop' }}
+                  push: true #${{ github.ref == 'refs/heads/develop' }}
```
romintomasetti: Revert this before merging.

.github/workflows/build.yml
```diff
@@ -43,7 +43,7 @@ jobs:
                 include:
                     - backend        : Cuda
                       compiler_family: gcc
-                      base_image     : nvidia/cuda:12.8.1-devel-ubuntu24.04
+                      base_image     : nvidia/cuda:12.9.0-devel-ubuntu24.04
```
romintomasetti, line 46: Our runner still has its driver pointing to `12.8`. I've `apt update` it but it's still on `12.8`.

Let's wait a bit.

maartenarnst: OK. Could also be worth it to try `sudo ubuntu-drivers install`.

PR #60, opened 2025-05-07
timer: rename 'reset' to 'start'

romintomasetti:
## Summary

We decided to rename the `reset` to `start`.

## Related to

- extracted from #43


maartenarnst (approved):
OK! All good indeed!

PR #69, opened 2025-05-09
ci(hip): run tests with our self-hosted runner

romintomasetti:
## Summary

Enable `HIP` testing with our self-hosted runner.


.github/workflows/build.yml
```diff
@@ -109,7 +110,7 @@ jobs:
               with:
                   context: .
                   platforms: ${{ matrix.platforms }}
-                  push: ${{ github.ref == 'refs/heads/develop' }}
+                  push: true #${{ github.ref == 'refs/heads/develop' }}
```
romintomasetti: Don't forget to revert this.

PR #75, opened 2025-05-15
callbacks(matchers): add event id matcher

romintomasetti:
## Summary

This PR introduces (yet another) matcher that matches events based on their `event_id` field, if any.

## Related to

- extracted from #43

### DIFF
--- a/include/kokkos-utils/callbacks/Events.hpp
+++ b/include/kokkos-utils/callbacks/Events.hpp
@@ -261,15 +261,22 @@ concept NamedEvent =
         { event.name } -> std::same_as<std::string&>;
     };
 
-//! Concept to constrain any event type that is one of the given event types.
-template <typename T, typename... EventTypes>
-concept EventOneOf = impl::type_list_contains_v<T, EventTypes...>;
+//! Concept to constrain any event type that has a member variable @c event_id.
+template <typename EventType>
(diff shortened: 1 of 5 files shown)


include/kokkos-utils/callbacks/Events.hpp
```diff
@@ -261,6 +261,12 @@ concept NamedEvent =
         { event.name } -> std::same_as<std::string&>;
     };
 
+//! Concept to constrain any event type that has a member variable @c event_id.
+template <typename EventType>
+concept IdentifiableEvent = Event<EventType> && requires (EventType& event) {
```
maartenarnst: Should we do actually `IndexedEvent`? To be consistent with `NamedEvent` and `EnqueuedEvent`?

maartenarnst:
```suggestion
concept IdentifiableEvent = Event<EventType> && requires (EventType event) {
```

romintomasetti: :+1:

romintomasetti: :+1:

romintomasetti (commented):
:green_circle:

maartenarnst (approved):

PR #77, opened 2025-05-15
callbacks(matchers): refactor `EventBeginEnd[Event]IdMatcher`

romintomasetti:
## Summary

This PR renames `EventBeginEndEventIdMatcher` to `EventBeginEndIdMatcher`, and refactors it to make it simpler and aligned with what `Kokkos` promises.

## Related to

- needs #75 
- extracted from #43


include/kokkos-utils/callbacks/EventBeginEndIdMatcher.hpp
```diff
@@ -0,0 +1,46 @@
...
+ * The @ref matcher can be used to match a specific begin event of type @p BeginEventType.
+ *
+ * @note The identifiable event type set is composed of begin and end event types. For a given begin event
+ *       type, there is only one end event type that @c Kokkos guarantees will potentially have the same
+ *       @c event_id as the begin event.
+ *       For instance, given a @ref BeginParallelForEvent event, only one @ref EndParallelForEvent event
+ *       should match.
+ */
+template <Matcher BeginMatcherType, BeginEvent BeginEventType> requires MatcherFor<BeginMatcherType, BeginEventType>
+struct EventBeginEndIdMatcher
```
maartenarnst, line 22: We need to see if we keep the two call operators. One challenge is the case where we'd want the start and stop to be the same event type. In this case, having two functions wouldn't work.

maartenarnst: OK. Sorry. I see better now. We impose it to be the paired event.

maartenarnst, line 22: We should think of whether we should rename it using "start" and "stop".

The `BeginParallelFor`, `BeginParallelReduce`, `BeginParallelScan` are "begin" events in `Kokkos`. For these events, `Kokkos` profiling uses a `BeginFunction`. On our side we have a `BeginEvent` concept.

And so here we're generalising to many more events. And `Start` and `Stop` would be a way to suggest that we're not aligned with "Begin" and "End" from `Kokkos` here.

romintomasetti: The role of the event ID is nothing else but tagging two events of the same nature, one beginning and one ending.

maartenarnst: OK. Got it! I hadn't thought of that we also want there to be an `event_id` member. I understand better the note now too. 👍

include/kokkos-utils/callbacks/EventBeginEndIdMatcher.hpp
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * @brief Match a begin event and its corresponding end event based on their @c event_id.
+ *
+ * The @ref matcher can be used to match a specific begin event of type @p BeginEventType.
+ *
+ * @note The begin and end event types, constrained by the @ref BeginEvent and @ref EndEvent concepts,
+ *       respectively, have an @c event_id member. For a given event of a begin event type, there is
+ *       only one event of the corresponding end event type that will potentially have the same
+ *       @c event_id as the begin event. For instance, given a @ref BeginParallelForEvent
+ *       event, only one @ref EndParallelForEvent event should match.
```
maartenarnst, line 19: I have updated a bit the sentences. I removed the sentence that said that "Kokkos guarantees that ...". Because in the case of our manager, it's our manager that sets the event ids. And with `kokkos-tools`, it would be the tool that sets the event ids.

include/kokkos-utils/callbacks/Events.hpp
```diff
@@ -276,7 +276,24 @@ concept EnqueuedEvent = Event<EventType> && requires (EventType event) {
...
+struct PairedEventType;
 
+#define PAIRED_EVENT_TYPE(__begin_eventtype__, __end_eventtype__) \
+    template <> \
+    struct PairedEventType<__begin_eventtype__> { using type = __end_eventtype__; };
+
+PAIRED_EVENT_TYPE(BeginParallelForEvent,    EndParallelForEvent)
+PAIRED_EVENT_TYPE(BeginParallelReduceEvent, EndParallelReduceEvent)
+PAIRED_EVENT_TYPE(BeginParallelScanEvent,   EndParallelScanEvent)
+PAIRED_EVENT_TYPE(BeginFenceEvent,          EndFenceEvent)
```
maartenarnst: Here, I've limited the pairs to the pairs that we're currently using with `BeginEndEventIdMatcher`.

romintomasetti: I wouldn't do that. I've explicitly avoided the name `BeginEndPair` to allow `PairedEvent` to embrace a broader goal.

maartenarnst: OK. Added them back!

maartenarnst (approved):
🟢

PR #78, opened 2025-05-18
callbacks(matchers): remove `matcher` member from `EventQueueMatcher`

romintomasetti:
## Summary

This PR removes the inner matcher. We rather envision to use some `And<MatcherType1, MatcherType2>` instead.

## Related to

- extracted from #43


maartenarnst (approved):
Ok! That’s better indeed to extract it!

PR #84, opened 2025-05-22
tests(fixture): introduce an execution space instance test fixture

romintomasetti:
### Summary

This fixture is useful to ensure that a test runs with a new (possibly unique) execution space instance.

I've explicitly added it to the `include` directory because we clearly want to ship it.

### Related

- will be useful in #43

### DIFF
--- a/include/kokkos-utils/tests/fixtures/ExecutionSpaceInstanceFixture.hpp
+++ b/include/kokkos-utils/tests/fixtures/ExecutionSpaceInstanceFixture.hpp
@@ -1,0 +1,22 @@
+#ifndef KOKKOS_UTILS_TESTS_HELPERS_HPP
+#define KOKKOS_UTILS_TESTS_HELPERS_HPP
+
+#include "kokkos-utils/concepts/ExecutionSpace.hpp"
+
+namespace Kokkos::utils::tests
+{
+
+//! Fixture class that creates a new execution space instance.
+template <Kokkos::utils::concepts::ExecutionSpace Exec>
+class ExecutionSpaceInstanceFixture
+{
+public:
+    ExecutionSpaceInstanceFixture() : exec(Kokkos::Experimental::partition_space(Exec{}, 1)[0]) {}
+
+protected:
+    Exec exec;
+};
+
+} // namespace Kokkos::utils::tests
+
+#endif // KOKKOS_UTILS_TESTS_HELPERS_HPP
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,8 @@ target_sources(
             include/kokkos-utils/impl/type_list.hpp
             include/kokkos-utils/impl/type_traits.hpp
 
+            include/kokkos-utils/tests/fixtures/ExecutionSpaceInstanceFixture.hpp
+
             include/kokkos-utils/timer/Duration.hpp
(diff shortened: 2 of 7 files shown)


include/kokkos-utils/tests/fixtures/ExecutionSpaceInstanceFixture.hpp
```diff
@@ -0,0 +1,22 @@
+#ifndef KOKKOS_UTILS_TESTS_HELPERS_HPP
+#define KOKKOS_UTILS_TESTS_HELPERS_HPP
+
+#include "kokkos-utils/concepts/ExecutionSpace.hpp"
+
+namespace Kokkos::utils::tests
+{
+
+//! Fixture class that creates an new execution space instance.
```
maartenarnst:
```suggestion
//! Fixture class that creates a new execution space instance.
```

include/kokkos-utils/tests/fixtures/ExecutionSpaceInstanceFixture.hpp
```diff
@@ -0,0 +1,22 @@
...
+template <Kokkos::utils::concepts::ExecutionSpace Exec>
+class ExecutionSpaceInstanceFixture : public virtual ::testing::Test
+{
+public:
+    ExecutionSpaceInstanceFixture() : exec(Kokkos::Experimental::partition_space(Exec{}, 1)[0]) {}
+
+protected:
+    Exec exec {};
```
maartenarnst:
```suggestion
    Exec exec;
```
(if we'd want to signal that it's the default constructor that is constructing this member)

include/kokkos-utils/tests/fixtures/ExecutionSpaceInstanceFixture.hpp
romintomasetti: :heavy_check_mark:

include/kokkos-utils/tests/fixtures/ExecutionSpaceInstanceFixture.hpp
romintomasetti: :heavy_check_mark:

include/kokkos-utils/tests/fixtures/ExecutionSpaceInstanceFixture.hpp
```diff
@@ -0,0 +1,22 @@
...
+#define KOKKOS_UTILS_TESTS_HELPERS_HPP
+
+#include "kokkos-utils/concepts/ExecutionSpace.hpp"
+
+namespace Kokkos::utils::tests
+{
+
+//! Fixture class that creates an new execution space instance.
+template <Kokkos::utils::concepts::ExecutionSpace Exec>
+class ExecutionSpaceInstanceFixture : public virtual ::testing::Test
```
maartenarnst: We'll need to think a bit about how to deal with the dependency on `gtest` that we create here.

romintomasetti: It's not needed actually. Let's get rid of it.

maartenarnst: OK. Best solution indeed!

maartenarnst (approved):
🟢 !

PR #85, opened 2025-05-23
callbacks: `BeginEndTimerListener` uses paired begin/end event types

romintomasetti:
In #43, we used a "paired event" for the `BeginEndTimerListener`, but we overlooked deleting the end event from the template arguments. We thus correct it here.


maartenarnst (approved):
OK! 🟢 👀

PR #87, opened 2025-05-31
callbacks: rename/move manager fixture

romintomasetti:
## Summary

I was trying to use `Kokkos::utils::callbacks::ManagerTestFixture` from `kokkos-utils/callbacks/Helpers.hpp` but came across:
* the header depends on `gmock/gmock.h`
* it derives from `::testing::Test`, thus depends on `GTest`

This PR breaks these dependencies.


include/kokkos-utils/tests/fixtures/callbacks/Manager.hpp
```diff
@@ -0,0 +1,18 @@
...
+{
+//! Fixture for initializing and finalizing @ref Kokkos::utils::callbacks::Manager.
+class Manager
+{
+public:
+    Manager()  { Kokkos::utils::callbacks::Manager::initialize(); }
+    ~Manager() { Kokkos::utils::callbacks::Manager::finalize(); }
+};
+
+} // namespace Kokkos::utils::tests::fixtures::callbacks
```
maartenarnst: I wonder if this is still a "fixture" now that it's no longer really related to tests. Does this class now look more like a "scoped" class?

If that's the case, it's probably true for the `ExecutionSpaceInstance` too. We could rename the `tests/fixtures` as  `tests/scoped`. And we could modify a bit this class like:  

```suggestion
namespace Kokkos::utils::tests::scoped::callbacks
{
//! Scoped class for initializing and finalizing @ref Kokkos::utils::callbacks::Manager.
struct KOKKOS_ATTRIBUTE_NODISCARD Manager
{
    Manager()  { Kokkos::utils::callbacks::Manager::initialize(); }
    ~Manager() { Kokkos::utils::callbacks::Manager::finalize(); }

    Manager& operator=(const Manager&) = delete;
    Manager(const Manager&)            = delete;
};

} // namespace Kokkos::utils::tests::scoped::callbacks
```

tests/callbacks/test_Manager.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include "kokkos-utils/callbacks/Helpers.hpp"
 #include "kokkos-utils/impl/type_traits.hpp"
+#include "kokkos-utils/tests/fixtures/callbacks/Manager.hpp"
```
maartenarnst: Perhaps this include could be provided directly in "kokkos-utils/callbacks/Helpers.hpp"?

romintomasetti: Nope. `kokkos-utils/callbacks/Helpers.hpp` should probably be about matchers only. And it's clearly not at the right place.

include/kokkos-utils/tests/fixtures/callbacks/Manager.hpp
romintomasetti: Done.

maartenarnst (approved):
🟢

tests/callbacks/test_Manager.cpp
maartenarnst: OK. Stay tuned :)

PR #91, opened 2025-07-03
callbacks(helpers): partial matchers

romintomasetti:
## Summary

This PR adds missing `DEFINE_EVENT_MATCHER` for `BeginDeepCopyEvent` and `EndDeepCopyEvent`.

It also introduces *partial matchers* that are especially useful for comparing events *partially*.

For instance, for allocation-related events, we generally don't want to check the pointer address.

### DIFF
--- a/include/kokkos-utils/callbacks/Events.hpp
+++ b/include/kokkos-utils/callbacks/Events.hpp
@@ -303,10 +303,14 @@ using paired_event_t = typename PairedEventType<EventType>::type;
 
 //! @name Stream operators.
 ///@{
-//! Helper function related to the implementation of the stream operators.
-template <Event EventType>
+/**
+ * @brief Helper function to remove the @c Kokkos::utils::callbacks:: prefix.
+ *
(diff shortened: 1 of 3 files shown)


include/kokkos-utils/callbacks/Events.hpp
```diff
@@ -303,10 +303,14 @@ using paired_event_t = typename PairedEventType<EventType>::type;
 
 //! @name Stream operators.
 ///@{
-//! Helper function related to the implementation of the stream operators.
-template <Event EventType>
+/**
+ * @brief Helper function related to the implementation of the stream operators that removes the @c Kokkos::utils::callbacks:: prefix.
```
maartenarnst:
```suggestion
 * @brief Helper function to remove the @c Kokkos::utils::callbacks:: prefix.
```
(best to keep the `that` right after the noun if you keep the subsentence with "that ...")

maartenarnst (dismissed):
🟢

include/kokkos-utils/callbacks/Events.hpp
romintomasetti: Done.

PR #92, opened 2025-07-05
Introduce a partial ordering for alloc descriptor.

maartenarnst:
This PR is a follow up for:
- #91  

The idea of this PR is to link the "partial" of the "partial matcher" to how C++ 20 allows us to define a weak or partial ordering for objects.

I.e., we overload the comparison operator of the `AllocDescriptor` with a custom comparison operator that does not take into account the pointer member. 

And then we define a new matcher that checks equivalence using the three-way comparison operator.

An advantage is that assertions become a bit shorter:

```
ASSERT_THAT(event, PartialMatcher<BeginDeepCopyEvent>{}(partial_match));
```

becomes

```
ASSERT_THAT(event, Equivalent(partial_match));
```

@romintomasetti, what do you think?

### DIFF
--- a/include/kokkos-utils/callbacks/Events.hpp
+++ b/include/kokkos-utils/callbacks/Events.hpp
@@ -106,6 +106,15 @@ struct AllocDescriptor
     uint64_t size = 0;
 
     bool operator==(const AllocDescriptor&) const = default;
+
+    std::partial_ordering operator<=>(const AllocDescriptor& other) const
+    {
+        if (kpsh != other.kpsh) {
+            return std::partial_ordering::unordered;
+        } else {
+            return std::tie(name, size) <=> std::tie(other.name, other.size);
+        }
+    }
 };
 
 //! Allocate-data event associated with @c Kokkos::Tools::Experimental::EventSet::allocate_data.
@@ -131,6 +140,8 @@ struct BeginDeepCopyEvent
     AllocDescriptor src {};
 
     bool operator==(const BeginDeepCopyEvent&) const = default;
+
+    std::partial_ordering operator<=>(const BeginDeepCopyEvent& other) const = default;
 };
 
 //! End-deep-copy event associated with @c Kokkos::Tools::Experimental::EventSet::end_deep_copy.
--- a/include/kokkos-utils/callbacks/Helpers.hpp
+++ b/include/kokkos-utils/callbacks/Helpers.hpp
@@ -184,6 +184,9 @@ struct PartialMatcher<BeginDeepCopyEvent>
     }
 };
 
+//! Defines a matcher @c Equivalent() to match an argument that is equivalent to @c val in the sense of a partial or weak ordering.
+MATCHER_P(Equivalent, val, "") { const auto cmp = arg <=> val; return cmp == std::partial_ordering::equivalent; }
+
 } // namespace Kokkos::utils::callbacks
 
 #endif // KOKKOS_UTILS_CALLBACKS_HELPERS_HPP
--- a/tests/callbacks/test_Helpers.cpp
+++ b/tests/callbacks/test_Helpers.cpp
@@ -91,4 +91,48 @@ TEST(PartialMatcher, BeginDeepCopyEvent)
     ASSERT_THAT(event, ::testing::Not(PartialMatcher<BeginDeepCopyEvent>{}(partial_no_match_src)));
 }
 
+//! @test Check that @ref Kokkos::utils::callbacks::Equivalent works as expected for @ref Kokkos::utils::callbacks::AllocDescriptor.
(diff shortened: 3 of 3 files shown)


romintomasetti:
The fact that the assertion becomes shorter does not hold at all.
I could have made a function like
```c++
template <typename T>
auto Equivalent(T arg) {
    return PartialMatcher<T>{}(std::move(arg));
}
```

romintomasetti:
I am not convinced it is the best way to achieve our goal here.

We want to have a partial matcher that is able to explain which field does not match for quick understanding of why a test failed (for instance, the name).

Your solution seems not really helpful in that sense because I guess it will just say "it is not equivalent" but will not give any detail.

romintomasetti:
Thank you for the proposal anyway.

I have one concern regarding the new `Equivalent` matcher: in the current code base, our partial matcher is based on Google Test's field matchers, which provide granular diagnostics—when a match fails, the error message points out exactly which field(s) differ. This can be very helpful for debugging.

With the new approach, the matcher only reports that the objects are not equivalent (according to the partial ordering), but does not specify which fields are responsible for the mismatch or for the unordered result. This could make debugging test failures less straightforward.

To address this, I experimented with a custom matcher that still uses the partial ordering logic but, in the case of failure, inspects each relevant field and reports detailed differences. This combines the expressiveness of partial ordering with the diagnostic power of field matchers. See the attached code for an example implementation.

But I find this too much convoluted code for such a simple task. All in all I would not go for your proposal.

```c++
#include <gmock/gmock.h>
#include <string>
#include <sstream>
#include "kokkos-utils/callbacks/Events.hpp"

class AllocDescriptorEquivalentMatcher
    : public ::testing::MatcherInterface<const AllocDescriptor&> {
public:
    explicit AllocDescriptorEquivalentMatcher(const AllocDescriptor& expected)
        : expected_(expected) {}

    bool MatchAndExplain(const AllocDescriptor& actual,
                         ::testing::MatchResultListener* listener) const override {
        auto cmp = actual <=> expected_;
        if (cmp == std::partial_ordering::equivalent) {
            return true;
        }

        std::ostringstream os;
        if (actual.kpsh != expected_.kpsh) {
            os << "kpsh fields differ: actual.kpsh = " << actual.kpsh.name
               << ", expected.kpsh = " << expected_.kpsh.name << "; ";
        }
        if (actual.name != expected_.name) {
            os << "name fields differ: actual.name = " << actual.name
               << ", expected.name = " << expected_.name << "; ";
        }
        if (actual.size != expected_.size) {
            os << "size fields differ: actual.size = " << actual.size
               << ", expected.size = " << expected_.size << "; ";
        }
        *listener << os.str();
        return false;
    }

    void DescribeTo(::std::ostream* os) const override {
        *os << "is equivalent (by partial order) to expected AllocDescriptor";
    }

private:
    const AllocDescriptor expected_;
};

inline ::testing::Matcher<const AllocDescriptor&> DiagnosticEquivalent(const AllocDescriptor& expected) {
    return ::testing::MakeMatcher(new AllocDescriptorEquivalentMatcher(expected));
}
```

maartenarnst:
That's a good idea indeed to look at the error messages!

If I make one of the assertions fail, we get with the current approach

```
Value of: event
Expected: (is an object whose given field (is an object whose given field is an object whose given field is equal to "Cuda") and (is an object whose given field is equal to "my-ds") and (is an object whose given field is equal to 128)) and (is an object whose given field (is an object whose given field is an object whose given field is equal to "Cuda") and (is an object whose given field is equal to "my-src") and (is an object whose given field is equal to 128))
  Actual: BeginDeepCopyEvent: {src = "my-src" (Cuda, 0x7ffee2b9d8f0) -> dst = "my-dst" (Cuda, 0x7ffee2b9d8f0) of size 128} (of type Kokkos::utils::callbacks::BeginDeepCopyEvent), whose given field is AllocDescriptor: {name = "my-dst" (Cuda, 0x7ffee2b9d8f0) of size 128} (of type Kokkos::utils::callbacks::AllocDescriptor), whose given field is "my-dst"
```

vs with the proposed approach:

```
Value of: event
Expected: equivalent (val: BeginDeepCopyEvent: {src = "my-src" (Cuda, 0) -> dst = "my-ds" (Cuda, 0) of size 128})
  Actual: BeginDeepCopyEvent: {src = "my-src" (Cuda, 0x7ffee2b9d8f0) -> dst = "my-dst" (Cuda, 0x7ffee2b9d8f0) of size 128} (of type Kokkos::utils::callbacks::BeginDeepCopyEvent)
```

And so the current approach gives a message that's a bit longer, but it indeed identifies where the problem is.

To me, it's fine either way! Let's let it rest a bit and decide on Monday ...

PR #95, opened 2025-08-28
docker(benchmark): update sha

romintomasetti:
## Summary

We're still using our fork, and I guess @maartenarnst recently force pushed on the branch.


maartenarnst (approved):
🟢 OK! That is the latest sha indeed!

PR #118, opened 2026-01-23
cicd: various updates

romintomasetti:
Mostly cmake.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,8 +1,8 @@
 {
-    "version"   : 3,
+    "version": 10,
     "configurePresets" : [
         {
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -170,7 +170,7 @@ jobs:
               with:
                   context: .
                   platforms: ${{ matrix.platforms }}
-                  push: ${{ github.ref == 'refs/heads/develop' }}
+                  push: true #${{ github.ref == 'refs/heads/develop' }}
```
romintomasetti: To be reverted.

PR #135, opened 2026-03-26
docker: allow any clang/gcc compiler version install

romintomasetti:
This PR allows to specify the version of the `clang` or `gcc` compiler we want to install.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,7 +44,7 @@ jobs:
                   files: |
                       .github/**
                       cmake/presets/**
-                      docker/dockerfile.builder
+                      docker/*
                       requirements/**
                       version.json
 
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -175,16 +186,17 @@ jobs:
               with:
                   context: .
                   platforms: ${{ matrix.platforms }}
-                  push: ${{ github.ref == 'refs/heads/develop' }}
+                  push: true #${{ github.ref == 'refs/heads/develop' }}
```
romintomasetti: Don't forget to revert this.

maartenarnst (dismissed):
Okay! Nice!

.github/workflows/build.yml
romintomasetti: Done.

PR #146, opened 2026-05-30
Tests: fence exec on destruction in scoped helper

maartenarnst:
This PR adds a destructor with a fence to our scoped ExecutionSpace storage. The motivation is to avoid issues with circular dependencies causing memory leaks on HPX. Fencing on destruction is consistent with what Kokkos does for default instances.


romintomasetti (changes requested):
I'm not sure this is required. I though Kokkos instances would already fence on destruction.

Search https://github.com/search?q=repo%3Akokkos%2Fkokkos+%22fence+on+destruction%22&type=code reveals they all do.

maartenarnst:
Closing as this is not needed in this form. We could revisit it as a way to break circular dependencies in the particular case of HPX.

PR #148, opened 2026-06-15
fix(Cuda): workaround for ice in nvcc+gcc

maartenarnst:
Seen in:
- uliegecsm/graph-dispatching#476

Solution found by @romintomasetti.


maartenarnst:
All good! @romintomasetti

